PR #42, opened 2020-04-23
Fixed: InstantType is decoded incorrectly [#38]

### LINKED ISSUE #38: InstantType is decoded incorrectly
A value decoded InstantType differs from the one from encoded object.
This is an actual bug as the decoder does not compensate for the fact that Spark stores timestamps with microseconds precision., so decoded instants are way in the future.

piotrszul:
Fixes #38.


johngrimes:
Merged as part of #57.

PR #48, opened 2020-04-23
Remove unused encoder functionality and flatten structure

### LINKED ISSUE #37: Remove Bundles functionality from encoders
This is functionality that we do not use, and the tests are currently invalid as the test data is not valid R4 FHIR.

johngrimes:
Fixes #37.


johngrimes:
Merged as part of #57.

PR #65, opened 2020-05-12
decimal encoding does not preserve precision.

### LINKED ISSUE #39: decimal encoding does not preserve precision.
The spec (http://hl7.org/fhir/datatypes.html) say:

> The precision of the decimal value has significance:
> e.g. 0.010 is regarded as different to 0.01, and the original precision should be preserved
> Implementations SHALL handle decimal values in ways that preserve and respect the precision of the value as represented for presentation purposes

The way we store it now always converts precision to 4 decimal places. This is agains the specification which says that precision should be preserved and 1.0000 is different than 1.00. For currency amounts it adds unnecessary 0s e.g. 22.22 becomes 22.2000 and for some high precision values e.g. from observations drops some decimal places e.g. 0.0232323232 -> 0.0233.

johngrimes:
Fixes #39.


johngrimes:
This change alters the encoding method for decimal types - which means that it will require existing data to be re-imported.

Issue #76, opened 2020-05-25
URI value in grouping causes error


johngrimes:
Request:

```json
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "subjectResource",
      "valueCode": "Encounter"
    },
    {
      "name": "aggregation",
      "part": [
        {
          "name": "expression",
          "valueString": "count()"
        },
        {
          "name": "label",
          "valueString": "Number of encounters"
        }
      ]
    },
    {
      "name": "grouping",
      "part": [
        {
          "name": "expression",
          "valueString": "reverseResolve(Condition.encounter).code.coding.system"
        },
        {
          "name": "label",
          "valueString": "Code system"
        }
      ]
    }
  ]
}
```

Error:

```
Caused by: java.lang.ClassCastException: org.hl7.fhir.r4.model.UriType cannot be cast to org.hl7.fhir.r4.model.StringType
	at au.csiro.pathling.query.parsing.LiteralComposer.getFhirPathForType(LiteralComposer.java:32)
	at au.csiro.pathling.query.AggregateExecutor.buildDrillDown(AggregateExecutor.java:291)
	at au.csiro.pathling.query.AggregateExecutor.lambda$mapRowToGrouping$5(AggregateExecutor.java:241)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
	at au.csiro.pathling.query.AggregateExecutor.buildResponse(AggregateExecutor.java:212)
	at au.csiro.pathling.query.AggregateExecutor.execute(AggregateExecutor.java:124)
	... 31 common frames omitted
```

Issue #70, opened 2020-05-19
HTML rendering of resource shows incorrect Content-Type


johngrimes:
To reproduce:

1. Visit https://pathling.csiro.au/fhir/OperationDefinition/aggregate-2 using a browser.
2. Observe that the `Content-Type` in the "Response Headers" section is `application/fhir+xml;charset=utf-8`, while the "Response Body" is JSON.

If you click on the "HTML JSON" or "HTML XML" links, the `Content-Type` appears to be set correctly.

piotrszul:
This is an issue with: ```ca.uhn.fhir.rest.server.interceptor.ResponseHighlighterInterceptor```
It uses different mechanism to determine determine encoding for embedding  response in HTML and for outputting the supposed response Content-Type.  In particular the second one is determined by the  content of "Accept: " request header. ```html/xml``` is selected because it's (in Chrome) it's explicitly accepted with high priority.

piotrszul:
This would have to be fixed in HAPI (?) or maybe we could create copy of this interceptor.

johngrimes:
Thanks @piotrszul, perhaps we should create a new issue/PR with hapi-fhir?

johngrimes:
Re-tested on v3 branch - this has now been fixed.

PR #138, opened 2020-09-17
v3

### LINKED ISSUE #2: Multiple aggregations on elements in single request produce inflated counts
Aggregation such as `name.given.count()` and `identifier.count()` produce different results if run in separate queries and when run together in a single query. 

The cause of the issue is that since both expressions produce lists with multiple elements per resource, when run together the aggregation (`count()`) runs over the Cartesian product of the lists (per resource), resulting in inflated counts.

The ignored test `AggregateExecutorTest#queryMultipleCountAggregations` verifies the correct behaviour.

johngrimes:
This new release incorporates a number of changes:

- Move to the Spring Boot framework
- Upgrade to Spark 3
- Refactored FHIRPath class model
- Request caching
- Improved error handling and reporting
- Improved configuration
- Support for invoking the `aggregate` operation with a GET request
- Bug fixes and performance improvements

#### Breaking changes

- Changes have been made to the encoding of the Parquet files from FHIR - any pre-existing Parquet data will need to be re-created from the FHIR sources before use with Pathling 3.
- The format of the request to the `aggregate` operation has been simplified, removing the labels and flattening the structure of the parameters.
- Configuration has been completely reworked - see the [Configuration](https://pathing.csiro.au/docs/configuration.html) page for the new variable names.
- Pathling now uses Spark 3 - this change will need to be reflected in any Spark cluster installations used to host Pathling.
- The semantics of the `subsumes` function have changed, and may yield different results in Pathling 3. See
  [subsumes](https://pathling.csiro.au/docs/fhirpath/functions.html) for the new specification.
- Source URLs used with the `import` operation now need to be whitelisted, for security purposes. A list of valid prefixes can be configured, by default you can import from `file:///usr/share/staging`.

Resolves #2, #6, #70, #76, #88, #157, #167, #175, #179, #181, #182, #184, #185, #186, #196, #202, #212, #213, #218, #220, #229, #233.

Tasks:

- [x] Update all dependencies to their latest versions
- [x] Update licences and licence plugin configuration
- [x] Commit design updates
- [x] Create a description of the breaking changes within the release
- [x] Add configuration to force correct language detection for repository
- [x] Add config to GitHub Actions to publish JAR to GitHub Packages


johngrimes:
Here is a diagram representing the new FHIRPath class model:

fhirpath-class-model

/cc @piotrszul

fhir-server/src/main/java/au/csiro/pathling/aggregate/FreshAggregateExecutor.java
```diff
@@ -137,16 +138,16 @@ public AggregateResponse execute(@Nonnull final AggregateRequest query) {
...
       @Nonnull final Collection<FhirPath> filters, @Nonnull final Column idColumn) {
     final List<Dataset<Row>> datasets = groupings.stream()
         .map(grouping -> {
           // We need to remove any trailing null values from non-empty collections, so that
           // aggregations do not count non-empty collections in the empty collection grouping. We do
           // this by joining the distinct set of resource IDs to the dataset using an outer join,
           // where the value is not null.
-          final Dataset<Row> distinctIds = grouping.getDataset().select(idColumn).distinct();
-          return join(grouping.getDataset(), idColumn, distinctIds,
+          return join(grouping.getDataset(), idColumn, inputContext.getDataset(),
```
johngrimes, line 150: @piotrszul I made the change you suggested, and it works and fixes the problem of the duplicate evaluation of paths.

PR #140, opened 2020-09-18
Refactor where function

### LINKED ISSUE #116: Refactor where function
Reimplement the where function as part of the refactored design of the v3 implementation.

Tasks:

- [x] Make join helpers support joining of grouped datasets
- [x] Join from distinct identities to argument dataset filtered of null values
- [x] Investigate why there seem to be multiple stages loading the same Parquet file, with the same filters

johngrimes:
Some changes made as part of this:

- Introduction of the NonLiteralPath.copy function, which is needed for both the creation of the where result, and also for creating the $this context
- Made DateLiteralPath capable of remembering the format which it was created from, so it is the same when used in drill-down expressions
- Changed AggregateFunction to preserve the input value column when there is a $this context
- Added tracking and preservation of $this and foreign resource contexts
- Enhanced query helpers to work with grouping columns
- Changed the language level to 11, and moved to Java 11 within GitHub Actions
- Added caching to ResourceReader, and made it possible to independently configure the cache size of each of the different cache types
- Upgraded to Spark 3.0.1
- Upgraded to HAPI 5.1.0

Resolves #116.


johngrimes:
I have come across a performance problem within this implementation: #142 

I don't have any ideas on how it might be fixed yet - any ideas from @piotrszul would be appreciated.

fhir-server/src/main/java/au/csiro/pathling/Configuration.java
```diff
@@ -296,7 +299,20 @@
      * Controls the maximum number of cache entries held in memory.
      */
     @NotNull
-    private long maxEntries;
+    @Min(0)
+    private long aggregateRequestCacheSize;
```
piotrszul, line 303: There is a bit of inconsistency of how the primitive types are represented here, in particular in case of @NotNull fields.
E.g. : 
```
@NotNull
    @Min(0)
    private Integer socketTimeout;
```

vs 

```
    @NotNull
    @Min(0)
    private long searchPageCacheSize;
```
I think in case of @NonNull fields it make sense to represent them as primitive types.

fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
```diff
@@ -70,8 +71,10 @@ protected QueryExecutor(@Nonnull final Configuration configuration,
   @Nonnull
   protected ParserContext buildParserContext(@Nonnull final ResourceType resourceType) {
     final ResourcePath inputContext = buildInputContext(resourceType);
-    return new ParserContext(inputContext, Optional.empty(), fhirContext, sparkSession,
-        resourceReader, terminologyClient, terminologyClientFactory);
+    check(inputContext.getIdColumn().isPresent());
```
piotrszul, line 74: This seem's to be outside of control of the user/caller and a violation most likely indicates a bug in the core rather than an illegal argument, so perhaps could be better expressed with `assert`?

fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
```diff
@@ -82,6 +85,12 @@ private ResourcePath buildInputContext(@Nonnull final ResourceType resourceType)
...
 
+  /**
+   * Joins the datasets in a list together, using their resource identity columns.
+   * <p>
+   * Note that this should only be used in contexts where the input context is a single resource -
+   * it does not support joining using grouping columns.
+   */
   @Nonnull
   protected static Dataset<Row> joinExpressions(@Nonnull final List<FhirPath> expressions) {
     check(!expressions.isEmpty());
```
piotrszul, line 96: Add an error message?

piotrszul, line 96: Add an error message?

fhir-server/src/main/java/au/csiro/pathling/aggregate/FreshAggregateExecutor.java
```diff
@@ -78,7 +77,7 @@ public FreshAggregateExecutor(@Nonnull final Configuration configuration,
   @Nonnull
   @Override
   public AggregateResponse execute(@Nonnull final AggregateRequest query) {
-    log.info("Executing request");
+    log.info("Executing request: " + query);
```
piotrszul, line 80: More efficient to to use: 
```
 log.info("Executing request: {}",  query);
 ```

File: fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java [hunk 5]
```diff
@@ -62,34 +62,36 @@
...
-   * <p>
-   * This is required for reverse reference resolution, where we need to get to the resource to be
-   * joined to the source of the reverse resolve operation.
-   *
-   * @return A {@link ResourceDefinition}
-   */
-  @Nonnull
-  Optional<ResourceDefinition> getOriginType();
+  static Optional<Column> findIdColumn(@Nonnull final Object... inputs) {
+    return Stream.of(inputs)
```
piotrszul, line 89: More as a curiosity - the first two lines can be also written as:
```
    return Stream.of(inputs)
        .filter(FhirPath.class::isInstance)
        .map(FhirPath.class::cast)
```

fhir-server/src/main/java/au/csiro/pathling/fhirpath/NonLiteralPath.java
```diff
@@ -72,8 +87,29 @@ protected NonLiteralPath(@Nonnull final String expression, @Nonnull final Datase
...
   public abstract Optional<ElementDefinition> getChildElement(@Nonnull final String name);
 
+  /**
+   * Gets a this {@link Column} from any of the inputs, if there is one.
+   *
+   * @param inputs a collection of objects
+   * @return a {@link Column}, if one was found
+   */
+  @Nonnull
+  public static Optional<Column> findThisColumn(@Nonnull final Object... inputs) {
```
piotrszul, line 106: Seems to be very similar to `FhirPath.findIdColumn`. Perhaps both can be refactored to use a helper method.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/UntypedResourcePath.java
```diff
@@ -56,13 +57,15 @@ protected UntypedResourcePath(@Nonnull final String expression,
...
    * @return a shiny new UntypedResourcePath
    */
   public static UntypedResourcePath build(@Nonnull final String expression,
       @Nonnull final Dataset<Row> dataset, @Nonnull final Optional<Column> idColumn,
-      @Nonnull final Column valueColumn, final boolean singular, @Nonnull final Column typeColumn,
+      @Nonnull final Column valueColumn, final boolean singular,
+      @Nonnull final Optional<Column> thisColumn, @Nonnull final Column typeColumn,
       @Nonnull final Set<ResourceType> possibleTypes) {
 
     final String hash = Integer.toString(Math.abs(dataset.hashCode()), 36);
```
piotrszul, line 71: `36` seems to be a magic number. Maybe `java.lang.Character.MAX_RADIX` instead?

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/BooleanPath.java
```diff
@@ -73,7 +66,7 @@ public BooleanPath(@Nonnull final String expression, @Nonnull final Dataset<Row>
   }
 
   @Override
```
piotrszul, line 68: Missing `@Notnull` annotation

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DatePath.java
```diff
@@ -42,20 +43,12 @@
...
-   * @param dataset A {@link Dataset} that can be used to evaluate this path against data
-   * @param idColumn A {@link Column} within the dataset containing the identity of the subject
-   * resource
-   * @param valueColumn A {@link Column} within the dataset containing the values of the nodes
-   * @param singular An indicator of whether this path represents a single-valued collection
-   * @param fhirType The FHIR datatype for this path, note that there can be more than one FHIR
-   * type
-   */
-  public DatePath(@Nonnull final String expression, @Nonnull final Dataset<Row> dataset,
+  protected DatePath(@Nonnull final String expression, @Nonnull final Dataset<Row> dataset,
```
piotrszul, line 46: Maybe protected constructors for abstract classes should have javadocs documentation? (that applies to all XXXPath)

fhir-server/src/main/java/au/csiro/pathling/Configuration.java
johngrimes: I can't work out if you still need `@NotNull` if the type is primitive.

I think I'll just standardise on primitive types and keep the `@NotNull` annotations...

johngrimes: Actually, that doesn't seem to work.

If I change `socketTimeout` to an `int`, then remove it from `application.yml`, it seems to load without error but assign `0` as the value. This is not the behaviour that we want, we want it to throw an error in this case.

I think this might have been the reason that I used the boxed primitive for that value in the first place.

johngrimes: I can confirm that it all works as desired when using the boxed primitives.

So maybe the action here is to standardise on the boxed primitives, including those `long` values that I have added as part of this PR.

fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
johngrimes: I actually have a `checkPresent` method which I should be using here, so I'll use that. It uses `Optional.orElseThrow()`.

We were previously using assertions quite liberally, but I wanted to move away from that. The reason for that was that we don't always have control over the execution environment that Pathling is deployed within, and assertions are not always enabled within the JVM.

johngrimes: What I have done is change the type of exception thrown by `Preconditions.check` to be an `AssertionError`, to better fit as a replacement for `assert`.

I've also added this new method that throws an `IllegalArgumentException`, and also accepts an error message:

```java
  /**
   * Ensures the truth of an expression, throwing an {@link IllegalArgumentException} with the
   * supplied message if it does not evaluate as true.
   *
   * @param expression The expression that should be true
   * @param errorMessage The message to use if an error is thrown
   */
  public static void checkArgument(final boolean expression, @Nonnull final String errorMessage) {
    if (!expression) {
      throw new IllegalArgumentException(errorMessage);
    }
  }
```

fhir-server/src/main/java/au/csiro/pathling/Configuration.java
johngrimes: See 5ad8d4094a18c8c18f84aec8fe3fa139338482fc.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/ReverseResolveFunction.java
```diff
@@ -48,27 +48,28 @@ public FhirPath invoke(@Nonnull final NamedFunctionInput input) {
 
     // Check that the argument types include the input type.
     final Set<ResourceType> argumentTypes = ((ReferencePath) argument).getResourceTypes();
-    final ResourceType inputType = ((ResourcePath) inputPath).getResourceType();
+    final ResourceType inputType = inputPath.getResourceType();
     checkUserInput(argumentTypes.contains(inputType),
         "Reference in argument to reverseResolve does not support input resource type: "
             + expression);
 
     // Do a left outer join from the input to the argument dataset using the reference field in the
```
piotrszul, line 56: This most likely is OK but the comment is confusing ...

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/WhereFunction.java
```diff
@@ -0,0 +1,95 @@
...
+      final DatasetWithColumns datasetWithColumns = joinOnColumns(distinctIds, groupingColumns,
+          argumentDataset, groupingColumns, Optional.of(argumentTrue), JoinType.LEFT_OUTER);
+      dataset = datasetWithColumns.getDataset();
+      input.getContext().setGroupingColumns(datasetWithColumns.getColumns());
+      idColumn = Optional.empty();
+    } else {
+      final Column argumentId = checkPresent(argumentPath.getIdColumn());
+      final String hash = randomShortString();
+      final String idColumnName = hash + ID_COLUMN_SUFFIX;
+      Dataset<Row> distinctIds = argumentDataset.withColumn(idColumnName, argumentId);
```
piotrszul, line 81: Maybe ??? 
```
Dataset<Row> distinctIds = argumentDataset.select(argumentId.alias(idColumnName)).distinct()
```

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/BooleanLiteralPath.java
```diff
@@ -85,5 +84,21 @@ public boolean isComparableTo(@Nonnull final Class<? extends Comparable> type) {
   public Optional<BooleanType> getValueFromRow(@Nonnull final Row row, final int columnNumber) {
     return BooleanPath.valueFromRow(row, columnNumber);
   }
-  
+
+  @Nonnull
+  @Override
+  public BooleanLiteralPath copy(@Nonnull final String expression,
```
piotrszul, line 90: This seem to be (almost) the same for all literals. Perhaps it could be refactored to LiteralPath to avoid code duplication.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/DateLiteralPath.java
```diff
@@ -53,28 +56,38 @@ public static DateLiteralPath fromString(@Nonnull final String fhirPath,
     check(context.getIdColumn().isPresent());
     final String dateString = fhirPath.replaceFirst("^@", "");
     java.util.Date date;
+    DateLiteralFormat format;
     // Try parsing out the date using the three possible formats, from full (most common) down to
     // the year only format.
     try {
       date = DatePath.getFullDateFormat().parse(dateString);
```
piotrszul, line 63: SimpleDataFormat is not thread safe. Since we use a singleton here (per spark executor) we might run into synchronisation issues with muti-threaded executors. 
I think the solution maybe to use thread local rather than static variable.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -141,81 +119,137 @@ public static IdAndValue updateGroupingColumns(@Nonnull final ParserContext cont
...
   /**
-   * @param left A {@link Dataset}
-   * @param leftId The ID {@link Column} in the left Dataset
-   * @param right Another Dataset
-   * @param rightId The ID column in the right Dataset
-   * @param joinType A {@link JoinType}
-   * @return A new {@link Dataset}
+   * Joins a {@link Dataset} to another Dataset, using resource identity columns.
+   * <p>
+   * This should not be used in contexts where there may be grouping columns.
```
piotrszul, line 196: There does not seem to be check for this (no grouping columns) anywhere.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -141,81 +119,137 @@ public static IdAndValue updateGroupingColumns(@Nonnull final ParserContext cont
...
-    return Arrays.asList(dataset.columns()).subList(0, numberOfColumns)
-        .stream()
-        .map(dataset::col)
-        .collect(Collectors.toList());
+  public static Dataset<Row> joinOnReference(@Nonnull final FhirPath left,
+      @Nonnull final FhirPath right, @Nonnull final JoinType joinType) {
+    @Nullable final Column reference = left.getValueColumn().getField("reference");
+    checkNotNull(reference);
+    final Column rightId = checkPresent(right.getIdColumn());
+    final Column joinCondition = reference.equalTo(rightId);
```
piotrszul, line 229: Maybe call the the version of `joinOnRefence` once `rightId` is obtained.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -306,11 +341,25 @@ public static DatasetWithColumns joinOnColumns(@Nonnull final Dataset<Row> left,
...
+   * Returns a list of the first N columns within a {@link Dataset}.
+   *
+   * @param dataset the Dataset to get the columns from
+   * @param numberOfColumns the number of columns to take
+   * @return a list of {@link Column} objects
+   */
+  @Nonnull
+  public static List<Column> firstNColumns(@Nonnull final Dataset<Row> dataset,
+      final int numberOfColumns) {
+    return Arrays.asList(dataset.columns()).subList(0, numberOfColumns)
```
piotrszul, line 359: Maybe: 
```
    return Stream.of(dataset.columns())
        .limit(numberOfColumns)
        .map(dataset::col)
        .collect(Collectors.toList());
```

fhir-server/src/test/java/au/csiro/pathling/test/builders/UntypedResourcePathBuilder.java
```diff
@@ -84,6 +88,12 @@ public UntypedResourcePathBuilder singular(final boolean singular) {
     return this;
   }
 
+  @Nonnull
+  public UntypedResourcePathBuilder thisColumn(final Column thisColumn) {
```
piotrszul, line 92: Is the nullability annotation missing here or do we have a conversion that if not annotated == @Nullable?

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
```diff
@@ -7,81 +7,92 @@
...
   public void whereOnResource() {
     final StructType encounterStruct = new StructType(new StructField[]{
         DataTypes.createStructField("id", DataTypes.StringType, true),
         DataTypes.createStructField("status", DataTypes.StringType, true)
     });
 
-    final Dataset<Row> dataset = new DatasetBuilder()
+    final Dataset<Row> inputDataset = new DatasetBuilder()
         .withIdColumn()
         .withStructTypeColumns(encounterStruct)
```
piotrszul, line 61: There is no row with NULL value.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
```diff
@@ -114,27 +124,33 @@ public void whereOnElement() {
         .withRow("Patient/abc3", "en")
         .withRow("Patient/abc3", "en")
         .withRow("Patient/abc3", "zh")
+        .withRow("Patient/abc4", "fr")
+        .withRow("Patient/abc4", "fr")
```
piotrszul, line 128: There is no row with NULL value.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
```diff
@@ -218,28 +184,33 @@ public void nullValuesAreNull() {
...
         .singular(false)
         .build();
 
     // Build an expression which represents the argument to the function.
-    final Column inputCol = inputExpression.getValueColumn();
-    final Column whereResult = functions.when(inputCol.equalTo("en"), null).otherwise(true);
-    final ElementPath argumentExpression = new ElementPathBuilder()
+    final Dataset<Row> argumentDataset = inputPath.getDataset()
+        .withColumn("value",
+            functions.when(inputPath.getValueColumn().equalTo("en"), null).otherwise(true));
```
piotrszul, line 199: I find it hard to understand what is being tested here. What would be the corresponding fhirpath expression?

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
```diff
@@ -329,4 +296,45 @@ public void throwsErrorIfArgumentNotSingular() {
...
+    final InvalidUserInputError error = assertThrows(
+        InvalidUserInputError.class,
+        () -> whereFunction.invoke(whereInput));
+    assertEquals(
+        "Input to where function cannot be a literal: true",
+        error.getMessage());
+  }
+
+  @Test
+  public void throwsErrorIfArgumentIsLiteral() {
```
piotrszul, line 322: I am not sure if it would be useful but speaking why:
```
xxxx.where(true) 
```
should be illegal?

fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
johngrimes: See b90fdb5850f3476a99eddff7e81979134cac03e5.

fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
johngrimes: See b90fdb5850f3476a99eddff7e81979134cac03e5.

piotrszul (commented):
Lot's of awesome code!

I think the only major issue is the unsynchronised access to a global SimpleDateFormat instances.

Some general comments/suggestions:

- perhaps cross-cutting concern such as caching with ResouceLoader or loggin/exception handling in SearchExecutor couild be cleaner implemented with AOP (e.g: https://niels.nu/blog/2017/spring-boot-aop.html)
- perhaps the type hierarchy for FhirPath couild be improved by refactoring  out common functionality for XXXXPath and XXXXLiteralPath to XXXXType interface, e.g:  IntegerType. BooleanType etc. There does seem to be any extra fields associated with the path type and also properties of being comparable or materializable are associated with path type regardless if it's an element or literal.

fhir-server/src/main/java/au/csiro/pathling/aggregate/FreshAggregateExecutor.java
johngrimes: See a47ccd7354c37bdf0f9381b8ddd727d22343dd60.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java
johngrimes: Now that I am reading about the difference between `instanceof` and `isInstance`, I am wondering how `instanceof` works at all for what we are using it for.

It sounds like `instanceof` is a compile-time only thing...

johngrimes: ... which can't be right, because it _does_ work to interrogate types at runtime. We have tests for this.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/NonLiteralPath.java
johngrimes: I'm going to hold off on doing this until #151 is fixed - as I have a feeling that `findIdColumn` is wrongly implemented anyway...

fhir-server/src/main/java/au/csiro/pathling/fhirpath/UntypedResourcePath.java
johngrimes: See c63665963d84106a6c678dcf884bb2e4b23a7501.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/BooleanPath.java
johngrimes: See 1a8e0286c680b3240077c7bab27f9bfe58b8f1f5.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DatePath.java
johngrimes: Let's aim to get the public APIs documented at this time - documenting more than that is a lot of work!

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/ReverseResolveFunction.java
johngrimes: See bbe14269aadd6b9290b98bcc353a3d8b4fb9cb4f.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java
johngrimes: Ok, learned scholar @dionmcm informs me that the two main differences are:

- `instanceof` provides a level of compile-time checking
- `isInstance` needs to be used when the class handle is not static

In light of this, I propose we use `instanceof` when we are dealing with static class identifiers, so that we can get the benefit of the compile-time checks.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/WhereFunction.java
johngrimes: I tried this and it fails a number of tests - I have not had much luck getting aliasing to work across joins in general.

This is why you will see a lot of `withColumn` stuff going on.

I don't completely understand why aliasing doesn't work, or why it only sometimes works.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/BooleanLiteralPath.java
johngrimes: I think we can actually move `FhirPath.copy` to `NonLiteralPath`, and add validation to make it illegal to invoke a function off a literal. This would remove this override from all subclasses of `LiteralPath`.

It is not expressly prohibited by the spec, but it doesn't make a lot of sense and I can't think of a scenario where it would be needed.

johngrimes: See 4a1eafea0b08b55765cd593f9c2344a613e9d579.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/DateLiteralPath.java
johngrimes: See 8f36d1434838ddced3e5e67c4e57d6bb77e3b7a5.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/BooleanLiteralPath.java
johngrimes: See 4a1eafea0b08b55765cd593f9c2344a613e9d579.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: We do check for the existence of the resource identity column, but we can't check for the grouping columns here as we don't have the context (which is where the grouping columns live).

I do agree that this stuff is a bit messy, but I haven't thought of a better way to do it as yet.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: See c97c0a716ac60c2410c9c924f5ba2e9afbc977bc.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
johngrimes: It's not illegal per the spec - but it adds quite a bit of complexity if we want to support it, and it's not very useful.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: See fe5f7a36c62cbd2af2e735054c832126aa439eb5.

fhir-server/src/test/java/au/csiro/pathling/test/builders/UntypedResourcePathBuilder.java
johngrimes: See 0c7f8ee500c313ee0ffa716fda34c44c6f68686a.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
johngrimes: See 9fb1a6e29c01238ea501bff04ba6a2cdd97bc3f2.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
johngrimes: See 9fb1a6e29c01238ea501bff04ba6a2cdd97bc3f2.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/WhereFunctionTest.java
johngrimes: Well, this is really just to cover off the case where we have `null` values in the argument, particularly mixed in with other values.

Umm, as for an expression... you would probably need a nested where. It is a bit contrived, but maybe something like this:

```
Patient.communication
  .language
  .where(
    $this.communication
      .language
      .where($this.coding.code != 'en') != 'en'
  )
```

johngrimes:
Thanks for the review @piotrszul!

I'll have a bit of a think about AOP and the suggested refactoring in relation to pulling out common aspects of literal and non-literal types.

PR #147, opened 2020-09-30
Aggregation functions cannot be nested

### LINKED ISSUE #6: Aggregation functions cannot be nested
Currently, the aggregation expression within the aggregate operation only works correctly if the aggregation function is at the end of the expression.

So you can't do something like this:

```
name.given.count() + name.family.count()
```

The correct result for this should be the number of given names of all patients within the grouping, plus the number of family names for all patients in the grouping.

johngrimes:
Fixes #6.


fhir-server/src/test/java/au/csiro/pathling/aggregate/AggregateQueryTest.java
```diff
@@ -385,6 +385,21 @@ void queryWithNestedAggregation() {
         response);
   }
 
+  @Test
```
piotrszul, line 388: This test is OK but the one above `queryWithNestedAggregation()` does not seem to be testing a nested aggregation of the same type so I am not sure if nested aggregation is covered for the case with grouping present.

fhir-server/src/main/java/au/csiro/pathling/aggregate/DrillDownBuilder.java
```diff
@@ -50,10 +50,10 @@ public DrillDownBuilder(@Nonnull final List<Optional<Type>> labels,
   /**
    * Generates the FHIRPath string.
    *
-   * @return A FHIRPath expression
+   * @return A FHIRPath expression, unless there are no groupings or filters
```
piotrszul, line 53: I guess this is the matter of convention but  rather than no drill down expression it could be also an empty one which would result in all resources being returned when queried.

piotrszul (commented):
I am wondering if it's necessary to introduce the optionality of drill-down expression given that the same semantics can be possibly achieved in a simpler way (both on the server and the client) with an empty ('') drill down expression?

fhir-server/src/main/java/au/csiro/pathling/aggregate/DrillDownBuilder.java
johngrimes: Resolved that we will only supply a drill down expression where one is needed. We will also enforce that empty expressions cannot be supplied to operations - this change will be made as part of #201.

fhir-server/src/test/java/au/csiro/pathling/aggregate/AggregateQueryTest.java
johngrimes: See eee8f3b.

Issue #158, opened 2020-10-13
Preservation of element order and identity


piotrszul:
This is extracted from #115:

There are two enhancements I think we need to make to the solution as part of this ticket:

Preservation of element identity
Preservation of element order

Given a collection of given names (e.g. Patient.name.given):

1. Thomas
2. Jude
3. Evelyn
4. Evelyn

We should be able to both preserve the order of these elements, and also preserve the fact that there are two "Evelyn" values. The preservation of order is important for functions such as `first()`, and the preservation of duplicate elements will be important for certain aggregation functions that we have not implemented yet (e.g. `mean()`).

One idea that we have talked about for achieving this is to create an "element identifier" column upon explode operations. It would need to be unique within a resource, so it would probably be created as a concatenation with parent element identifiers created from previous explode operations.

If we need to aggregate against each individual element, instead of just using the resource identifier, we can use both the resource identifier and the element identifier. This might be a way of fixing #136.

Order can then be enforced by sorting on resource identifier and element identifier. Perhaps we only need to explicitly order at the time of invocation of an order-sensitive function such as `first()`.

johngrimes:
Fixed by #191.

Issue #151, opened 2020-10-06
Non-grouped sub-expressions break aggregation


johngrimes:
Use of the following aggregation expression with one or more groupings:

```
count() = 12
```

Results in aggregation on resource identity, rather than on the grouping columns.

johngrimes:
There may be an opportunity to rationalise `findIdColumn` and `findThisColumn` as part of this ticket: see https://github.com/aehrc/pathling/pull/140#discussion_r499967896.

johngrimes:
The following aggregation expression causes a similar result: `true`

johngrimes:
Fixed by #191.

PR #191, opened 2020-11-24
Query optimisation and element identity

### LINKED ISSUE #142: Optimise boolean expression within where argument
The following expression is quite simple but performs exceptionally poorly. It also seems to use a very large amount of memory.

```
reverseResolve(Observation.subject)
    .where(
        $this.status = 'preliminary'
        or
        $this.status = 'final'
    ).status
```

piotrszul:
Fixes #142, #151 and #158.


johngrimes:
@piotrszul The next task is to merge `issue/158` into this branch and get the tests passing.

johngrimes:
@piotrszul Also, all `@Disabled` tests will need to be re-enabled and passing.

johngrimes:
@piotrszul I'm ready to merge this when you are happy with the changes I've made.

piotrszul:
@johngrimes   I have two suggestions:

in:

```
  @Nonnull
  public static Dataset<Row> join(@Nonnull final FhirPath left, @Nonnull final FhirPath right,
      @Nonnull final JoinType joinType) {
    final List<Column> leftColumns = new ArrayList<>();
    final List<Column> rightColumns = new ArrayList<>();
    leftColumns.add(left.getIdColumn());
    rightColumns.add(right.getIdColumn());

    // If the two paths are NonLiteralPaths, and they have $this columns, we need to add the element
    // ID columns into the join condition. This is to prevent too many rows being generated when
    // there are more rows than resources on either side of the join.
    final boolean nonLiteralJoin =
        left instanceof NonLiteralPath && right instanceof NonLiteralPath;
    if (nonLiteralJoin && ((NonLiteralPath) left).getThisColumn().isPresent()
        && ((NonLiteralPath) right).getThisColumn().isPresent()) {
      leftColumns.add(((NonLiteralPath) left).getThisColumn().get());
      rightColumns.add(((NonLiteralPath) right).getThisColumn().get());
    }

    return join(left.getDataset(), leftColumns, right.getDataset(), rightColumns, joinType);
  }
```

we do not actually need join on the entire $this column but only on the "eid" field of the structure (may possibly be more performant)

piotrszul:
and in:

```
  @Nonnull
  private Dataset<Row> joinGroupingsAndFilters(@Nonnull final Collection<FhirPath> groupings,
      @Nonnull final Collection<FhirPath> filters, @Nonnull final Column idColumn) {
    final List<Dataset<Row>> datasets = groupings.stream()
        .map(grouping -> {
          // We need to remove any trailing null values from non-empty collections, so that
          // aggregations do not count non-empty collections in the empty collection grouping. We do
          // this by joining the distinct set of resource IDs to the dataset using an outer join,
          // where the value is not null.
          final Dataset<Row> distinctIds = grouping.getDataset().select(idColumn).distinct();
          return join(grouping.getDataset(), idColumn, distinctIds,
              idColumn, grouping.getValueColumn().isNotNull(), JoinType.RIGHT_OUTER);
        }).collect(Collectors.toList());
    datasets.addAll(filters.stream()
        .map(FhirPath::getDataset)
        .collect(Collectors.toList()));

    return datasets.stream()
        .reduce((a, b) -> join(a, idColumn, b, idColumn, JoinType.LEFT_OUTER))
        .orElseThrow();
  }
```

If the base resource is the same then perhaps  we could join all the filters and grouping with a normal join (filtering for null values) and at the very end do do left join to  distinct resource IDs.

johngrimes:
@ametke I think we're ready for you to pick this up again (on the `v3` branch) and try it out with your data.

We still have a few tickets to close ahead of the official v3 release, but I think we're close enough to get your feedback.

PR #214, opened 2020-12-16
Remove labels from aggregate request

### LINKED ISSUE #178: Remove labels from aggregate request
We currently have a parameter for a label for each grouping and aggregation within the request for the [aggregate operation](https://pathling.csiro.au/docs/aggregate.html), even though we don't actually do anything with this information.

The request could be simplified by removing these labels, and it would remove the need to deal with parameters with parts.

It will also fix the currently undesirable behaviour of the request cache being invalidated when a label within the request changes.

johngrimes:
Resolves #178.

### DIFF
--- a/fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java
+++ b/fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java
@@ -10,7 +10,6 @@ import static au.csiro.pathling.utilities.Preconditions.checkUserInput;
 
 import au.csiro.pathling.errors.InvalidUserInputError;
 import java.util.List;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import lombok.Value;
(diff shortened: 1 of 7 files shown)


fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java
```diff
@@ -97,105 +105,13 @@ private static ResourceType getSubjectResource(@Nonnull final Parameters paramet
...
-              .stream()
-              .filter(part -> part.getName().equals("expression") &&
-                  part.getValue() instanceof StringType)
-              .findFirst()
-              .map(parameter -> parameter.getValue().toString())
-              .orElseThrow(() -> new InvalidUserInputError("Grouping must have expression"));
-          return new Grouping(label, expression);
-        })
+        .filter(
+            param -> param.getName().equals(parameter) && param.getValue() instanceof StringType)
```
piotrszul, line 112: Maybe 
```parameter.equals(param.getName())``` (to avoid potential null pointer exceptions).

File: fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java [hunk 2]
```diff
@@ -97,105 +105,13 @@ private static ResourceType getSubjectResource(@Nonnull final Parameters paramet
...
+        .map(aggregation -> aggregation.getValue().toString())
```
piotrszul, line 113: change 'aggregation' to 'expression'?

fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java
johngrimes: Fixed in 5a7d60235f3e49e57a0b5483db8b164e88e71187.

fhir-server/src/main/java/au/csiro/pathling/aggregate/AggregateRequest.java
johngrimes: Fixed in 5a7d60235f3e49e57a0b5483db8b164e88e71187.

PR #215, opened 2020-12-17
Added configuration of allowed URL sources for import.

### LINKED ISSUE #210: Add whitelisting of import sources
Currently you can instruct Pathling to import from arbitrary URLs, which could become a vector for malicious content.

This change will introduce the requirement to whitelist (through configuration) a set of URL prefixes which are allowable for use within the import operation.

Configuration would look something like this by default:

```yaml
pathling:
  import:
    allowableSources:
      - "file:///usr/share/staging"
```

piotrszul:
Resolves #210


johngrimes (approved):
Looks great, thanks @piotrszul!

PR #219, opened 2020-12-18
Fix resource ID encoding and reference resolution

### LINKED ISSUE #218: Resource identity incorrect
Currently, the encoders use `IdType.getValueAsString` to populate the ID element within the underlying dataset. This results in a "qualified identifier" being used within the `id` element, and it causes confusing behaviour when performing query operations directly on the `id` element. It also causes resource references to fail when the target has technical version metadata, as HAPI adds the version suffix on to the generated qualified identifier.

Two changes need to be made:

1. Modify the encoders such that only the "ID part" is populated into the `id` element.
2. Change resource reference resolution to convert relative literal references into raw identifiers before comparison.

The following things are NOT supported (along with issue references for the work required to support them, where this is planned):

1. References to individual technical versions of a resource (#216)
2. Logical references (via `identifier`) (#217)
3. References to contained resources
4. Absolute literal references

johngrimes:
Resolves #218.


johngrimes:
There is a caveat with this implementation, I've documented this in #222.

PR #250, opened 2021-01-28
SPIKE: Passing evaluation context back from term to invocation

piotrszul:
This is a spike that explores possibly for expression to select evaluation context for invocations.


johngrimes (commented):
We'll merge #249 in favour of this, but this pattern may be needed in the future.

johngrimes:
Closing in favour of #249.

PR #255, opened 2021-02-01
Implement iif function

### LINKED ISSUE #195: FHIRPath function: iif
See: [iif(criterion: expression, true-result: collection [, otherwise-result: collection]) : collection](http://hl7.org/fhirpath/#iifcriterion-expression-true-result-collection-otherwise-result-collection-collection)

johngrimes:
Resolves #195.


fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/IfFunction.java
```diff
@@ -0,0 +1,90 @@
...
+    // Build a new ElementPath based on the type of the literal `ifTrue` and `otherwise` arguments,
+    // and populate it with the dataset and calculated value column.
+    final String expression = expressionFromInput(input, NAME);
+    return ElementPath
+        .build(expression, dataset, inputPath.getIdColumn(), inputPath.getEidColumn(), valueColumn,
+            inputPath.isSingular(), inputPath.getForeignResource(), inputPath.getThisColumn(),
+            ifTrueType);
+  }
+
+  private FHIRDefinedType getFhirTypeForArgument(final FhirPath argument) {
```
piotrszul, line 77: @Nonnull missin

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPath.java
```diff
@@ -50,29 +51,54 @@ protected CodingLiteralPath(@Nonnull final Dataset<Row> dataset, @Nonnull final
...
+      final String result = unSingleQuote(component);
+      return StringLiteralPath.unescapeFhirPathString(result);
+    } else {
+      return component;
+    }
+  }
+
+  @Nonnull
+  private static String encodeComponent(@Nonnull final String component) {
+    if (component.matches("[',]")) {
```
piotrszul, line 85: This looks suspicious: 
Why ',' should indicate need for escaping.
Also probably should be component.matches(".*[',].*")

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPath.java
```diff
@@ -50,29 +51,54 @@ protected CodingLiteralPath(@Nonnull final Dataset<Row> dataset, @Nonnull final
...
       throw new IllegalArgumentException(
           "Coding literal must be of form [system]|[code] or [system]|[version]|[code]");
     }
     return new CodingLiteralPath(context.getDataset(), context.getIdColumn(), coding);
   }
 
+  @Nonnull
+  private static String decodeComponent(@Nonnull final String component) {
+    if (component.matches("^'(.*)'$")) {
+      final String result = unSingleQuote(component);
```
piotrszul, line 76: The regex can be probably precompiled for better perfomance.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/IfFunctionTest.java
```diff
@@ -0,0 +1,384 @@
...
+        .withEidColumn()
+        .withColumn(DataTypes.StringType)
+        .withRow("observation-1", makeEid(0), "foo")
+        .withRow("observation-2", makeEid(0), "bar")
+        .withRow("observation-3", makeEid(0), "bar")
+        .withRow("observation-4", makeEid(0), "foo")
+        .withRow("observation-4", makeEid(1), "bar")
+        .withRow("observation-5", makeEid(0), "bar")
+        .withRow("observation-5", makeEid(1), "bar")
+        .build();
```
piotrszul, line 119: I can see that we treat null values in the condition as 'false' but is that correct? 
Should we not rather preserve nulls?

piotrszul (commented):
Looks good to me except for a few comments inside the code.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/IfFunctionTest.java
johngrimes: I think that is the case for most FHIRPath functions, but the spec for this one says:

> If criterion is false or an empty collection, the function returns otherwise-result, unless the optional otherwise-result is not given, in which case the function returns an empty collection.

johngrimes: Note that we don't do the "optional" part currrently - I'm not really sure what the value of it is.

Doesn't leaving out the third argument just turn this into `where`?

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPath.java
johngrimes: Done.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPath.java
johngrimes: I've changed this to match the same pattern that is used for `RAW_CODING_COMPONENT`, which is any character except those used to delimit strings, arguments and coding components.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/IfFunction.java
johngrimes: Fixed.

johngrimes:
@piotrszul I found some cases that weren't covered by my original implementation - things like having `ResourcePath` or `UntypedResourcePath` in the arguments, or `BackboneElements`.

I have fixed this by introducing a method called FhirPath.mergeWith which knows how to create a new path which can be used to accommodate values from both the invoking path and a target path.

I've also added a bunch more test coverage.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java
```diff
@@ -47,4 +48,25 @@
...
+   * @param valueColumn a {@link Column} within the dataset containing the values of the nodes
+   * @param singular an indicator of whether this path represents a single-valued collection
+   * @param thisColumn for paths that traverse from the {@code $this} keyword, this column refers to
+   * the values in the collection
+   * @return the resulting new {@link NonLiteralPath}
+   */
+  NonLiteralPath mergeWith(FhirPath target, Dataset<Row> dataset, String expression,
+      Column idColumn, Optional<Column> eidColumn, Column valueColumn, boolean singular,
+      Optional<Column> thisColumn);
+
```
piotrszul, line 71: Add Nonnull annotations?

fhir-server/src/main/java/au/csiro/pathling/fhirpath/UntypedResourcePath.java
```diff
@@ -123,4 +124,23 @@ public UntypedResourcePath copy(@Nonnull final String expression,
...
+  @Nonnull
+  public NonLiteralPath mergeWith(@Nonnull final FhirPath target,
+      @Nonnull final Dataset<Row> dataset, @Nonnull final String expression,
+      @Nonnull final Column idColumn, @Nonnull final Optional<Column> eidColumn,
+      @Nonnull final Column valueColumn, final boolean singular,
+      @Nonnull final Optional<Column> thisColumn) {
+    if (target instanceof UntypedResourcePath && possibleTypes
+        .equals(((UntypedResourcePath) target).possibleTypes)) {
+      // Two UntypedResourcePaths can be merged together if they have the same set of possible
+      // types.
```
piotrszul, line 137: I think this could potentially be more lenient, that it the any combination of possibleTypes should be allowed by the results possibleTypes should be the union of both inputs.
Or per minimum it what should be allowed is merging when possibleTypes contain target.possibleTypes.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPathTest.java
```diff
@@ -0,0 +1,135 @@
...
+        .fromString(expression, inputContext);
+    final Coding literalValue = codingLiteralPath.getLiteralValue();
+    assertEquals("http://snomed.info/sct", literalValue.getSystem());
+    assertEquals("http://snomed.info/sct/32506021000036107/version/20201231",
+        literalValue.getVersion());
+    assertEquals("46,2", literalValue.getCode());
+
+    final String actualExpression = codingLiteralPath.getExpression();
+    assertEquals(expression, actualExpression);
+  }
```
piotrszul, line 127: But is this correct? 
If code value "46,2" should be quoted and escaped, should not he actualExpression have be:
" ... |`46\\,2"? 
Also to test escaping should the expression be " ... |`46\\,2" as well?

piotrszul (commented):
Looks good apart from a few comments in the code.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPathTest.java
johngrimes: I don't think commas need to be escaped in the context of a string literal: http://hl7.org/fhirpath/#string. They are, however, problematic _outside_ of a string literal.

If we had a Coding with a code that contained a comma that we wanted to encode as a Coding literal, we would need to quote the code to ensure that the parser didn't interpret it as the end of the Coding literal (and maybe the start of a new argument, if it was in the context of a function's arguments). The same goes for whitespace, parentheses and the pipe operator.

So `NEEDS_QUOTING` needs to be the minimal set of characters that indicate that the component will need to be quoted so that it doesn't fall foul of the patterns in `FhirPath.g4`.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java
johngrimes: I can't seem to be able to get IntelliJ to tell me about this sort of issue.

Maybe it's because this is an interface...

johngrimes: Actually, maybe this would have picked it up (in a kind of indirect way):

image

fhir-server/src/test/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPathTest.java
piotrszul: Ok, I guess I did not fully understand what needs to be quoted/escapes.
But do we have a test case for such a coding the needs escaping?

fhir-server/src/main/java/au/csiro/pathling/fhirpath/UntypedResourcePath.java
johngrimes: Interesting... as far as I can tell, we never actually use `possibleTypes`. And I think this is actually ok, any resources which don't match just don't get added to the resulting collection.

So you can pass anything you like as an argument to `ofType` - as long as it is a valid resource type code. It is not actually checked against the `possibleTypes` of the input path.

The spec doesn't really say that you need to check, from my reading: http://hl7.org/fhirpath/#oftypetype-type-specifier-collection

So I actually think we could remove `possibleTypes` altogether... unless I am missing something.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/FhirPath.java
johngrimes: Fixed in c6592b2.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/UntypedResourcePath.java
johngrimes: I've removed `possibleTypes` in e99e70a.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/literal/CodingLiteralPathTest.java
johngrimes: I have added more tests in 8ea952a.

PR #257, opened 2021-02-04
Implemented custom coder for IdType.

### LINKED ISSUE #222: Technical version information lost on deserialisation
The current implementation has a limitation relating to the deserialisation of technical versions.

Currently, if you import a resource with technical version information (i.e. `meta.versionId`), it will be available for query but it will not be included in the resource within search results, or any other operation that requires conversion of the data back into a FHIR resource.

This requires enhancement to the encoders to intellligently deserialise the `id` element based upon information within the resource `meta` element.

piotrszul:
IdType is not encoded on two columns:

 * 'id' column hold the abbreviated unversioned id.
 * 'id_versioned' column holds the full id with version, type etc.

Resolves #222.

### DIFF
--- a/encoders/src/main/scala/au/csiro/pathling/encoders/datatypes/IdCustomCoder.scala
+++ b/encoders/src/main/scala/au/csiro/pathling/encoders/datatypes/IdCustomCoder.scala
@@ -1,0 +1,59 @@
+/*
+ * This is a modified version of the Bunsen library, originally published at
+ * https://github.com/cerner/bunsen.
+ *
+ * Bunsen is copyright 2017 Cerner Innovation, Inc., and is licensed under
+ * the Apache License, version 2.0 (http://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * These modifications are copyright © 2018-2021, Commonwealth Scientific
+ * and Industrial Research Organisation (CSIRO) ABN 41 687 119 230. Licensed
+ * under the CSIRO Open Source Software Licence Agreement.
+ */
+
+package au.csiro.pathling.encoders.datatypes
+
+import org.apache.spark.sql.catalyst.expressions.objects.{Invoke, NewInstance, StaticInvoke}
+import org.apache.spark.sql.catalyst.expressions.{Expression, Literal}
+import org.apache.spark.sql.types._
+import org.apache.spark.unsafe.types.UTF8String
+import org.hl7.fhir.r4.model.IdType
+
+
+/**
+ * Custom coder for IdType.
+ * Represents IdType on two dataset columns:
+ * 'id' column hold the abbreviated unversioned id.
+ * 'id_versioned' column holds the full id with version, type etc.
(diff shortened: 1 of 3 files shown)


johngrimes:
@piotrszul Is this implementation tolerant of the absence of the `id_versioned` column, as would be the case in Parquet data that would have been imported prior to this change?

I have tested this change through invoking search on previously imported data, and it seems to work without error - but I thought I would double check.

piotrszul:
@johngrimes  not I do not think that it would/should work. I tested it with the encoders by trying to decode a dataset with 'versioned_id' removed and it does fail. 
Maybe a way to go around this would be to modify ResourcerReader to alias 'id' as 'id_versioned' in case it's not present in the parquet dataset.

piotrszul:
Or otherwise, we would have to reimport the data.

johngrimes:
I think the modification to `ResourceReader` sounds like it would be worthwhile, in order to make this change backwards-compatible for now.

Then I think we should create another issue to remove that logic (a breaking change), and roll that into the next major version release.

johngrimes (changes requested):
Closing this review, pending investigation into whether we can make this a backwards-compatible change.

piotrszul:
@johngrimes  So:
1) I have verified that this IS a breaking change - that is when I run search on the 1k_20190829 dataset with the old schema is does fail in the expected way.
2) I did try to add this code to ResoureReader to provide alias for 'id_versioned" and it does partially work:

```
    log.info("Loading resource {} from: {}", resourceType.toCode(), tableUrl);
    @Nullable final Dataset<Row> parquetResource = spark.read().parquet(tableUrl);
    checkNotNull(parquetResource);

    final Dataset<Row> resources = (QueryHelpers
        .hasColumn(parquetResource, functions.col("id_versioned")))
                                   ? parquetResource
                                   : parquetResource
                                       .withColumn("id_versioned", functions.col("id"));
    resources.cache();
    return resources; ```
    
However it appears that IdType is used in other places (in addition to resource 'id'), e.g. meta/versionId so the search still fail.

I do not think it's possible to create aliased for all possibly affected columns (is it ?)

So options I can see: 
1) use the other idea, to create unversioned id in ResourceReader, as in the code above. Since we cache the dataset after this operation it should not have adverse effect on performance (in theory).
2) Introduce some notion of schema version in encoders and and support for multiple version of parquet files possibly by providing mutilple versions of R4DataTypeMappings. So save the version with the warehouse and pick the right mappings to use for this version.
3) Or do some kind of hacking - maybe only add the '_versioned' if the name of the column is 'id' .

johngrimes:
We've decided to leave the implementation as is, and hold it over until the next major release.

The issue itself is non-critical, and we don't want to trigger a major version change for this fix alone.

johngrimes (approved):

johngrimes:
This was merged into `release/4.0.0`, and will go out in that release.

PR #259, opened 2021-02-04
dynamically set year in the licence template

dionmcm:


johngrimes:
@dionmcm Thanks for the pull request!

PR #307, opened 2021-05-12
Add operation and resource level authorisation

### LINKED ISSUE #282: Add operation and resource level authorisation
This change adds a number of new authorisation scopes, and changes the logic that is used to interpret scopes within tokens.

Here are the operations and the scopes that will be required to invoke them (when authorisation is enabled):

| Operation | Required scopes                                                                                     |
| --------- | --------------------------------------------------------------------------------------------------- |
| import    | (`user/*.write` OR `user/[resource type].write` for all specified resources) AND `operation:import` |
| aggregate | (`user/*.read` OR `user/[resource type].read` for all accessed resources) AND `operation:aggregate`  |
| search    | (`user/*.read` OR `user/[resource type].read` for all accessed resources) AND `operation:search`     |

The following requests are always allowed:

1. The read operation on the `OperationDefinition` resource;
2. Invocation of the [capabilities](http://hl7.org/fhir/r4/http.html#capabilities) operation, and;
3. The SMART configuration endpoint (`/.well-known/smart-configuration`).

An example import request:

POST `/fhir/$import`

```json
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "source",
      "part": [
        {
          "name": "resourceType",
          "valueCode": "Condition"
        },
        {
          "name": "url",
          "valueUrl": "s3://somebucket/Condition.ndjson"
        }
      ]
    },
    {
      "name": "source",
      "part": [
        {
          "name": "resourceType",
          "valueCode": "Encounter"
        },
        {
          "name": "url",
          "valueUrl": "s3://somebucket/Encounter.ndjson"
        }
      ]
    }
  ]
}
```

Would be authorised in the following scenarios:

| Scopes                                                           | Authorised? |
| ---------------------------------------------------------------- | ----------- |
| `user/*.write`                                                   | NO          |
| `user/Condition.write` `operation:import`                        | NO          |
| `user/*.write` `operation:import`                                | YES         |
| `user/Condition.write` `user/Encounter.write` `operation:import` | YES         |

An example aggregate request:

`POST /fhir/Patient/$aggregate`

```json
{
    "resourceType": "Parameters",
    "parameter": [
        {
            "name": "aggregation",
            "valueString": "count()"
        },
        {
            "name": "grouping",
            "valueString": "reverseResolve(Condition.subject).code.coding.where(subsumedBy(http://snomed.info/sct|73211009))"
        }
    ]
}
```

Would be authorised in the following scenarios:

| Scopes                                                           | Authorised? |
| ---------------------------------------------------------------- | ----------- |
| `user/*.read`                                                    | NO          |
| `user/Patient.read` `operation:aggregate`                        | NO          |
| `user/*.read` `operation:aggregate`                              | YES         |
| `user/Patient.read` `user/Condition.read` `operation:aggregate`  | YES         |

Note that either `user/*.read` or `user/Condition.read` is required, due to the call to `reverseResolve(Condition.subject)` within the grouping expression.

piotrszul:
Resolves #282


johngrimes:
There are three things left to do on this:

1. Investigate the possibility of sourcing `pathling.auth.authorizeUrl`, `pathling.auth.tokenUrl` and `pathling.auth.revokeUrl` from the discovery endpoint, allowing us to remove the configuration parameters
2. Get the existing test `au.csiro.pathling.test.integration.AuthorizationTest#smartConfiguration` passing
3. Add new tests which mock out the call to the discovery endpoint and verify that a JWT signed with that key is accepted by the implementation (and the negative case)

johngrimes:
@piotrszul I get the following error with the latest change:

java.lang.IllegalArgumentException: When allowCredentials is true, allowedOrigins cannot contain the special value "*" since that cannot be set on the "Access-Control-Allow-Origin" response header. To allow credentials to a set of origins, list them explicitly or consider using "allowedOriginPatterns" instead.

piotrszul:
> @piotrszul I get the following error with the latest change:
> 
> java.lang.IllegalArgumentException: When allowCredentials is true, allowedOrigins cannot contain the special value "*" since that cannot be set on the "Access-Control-Allow-Origin" response header. To allow credentials to a set of origins, list them explicitly or consider using "allowedOriginPatterns" instead.

Ah, yes. It is actually exactly what it says, but I think I might have misinterpreted the CORS specification a bit implementing it this way. 
The reason for this error is that I have added `Access-Control-Allow-Credentials: true` when security is enabled (which I assume is correct as we are sending credentials with the request), and when this is present `Access-Control-Allow-Origin` cannot be set to `*` at least not in the server response.  I thought it also implies it should not allowed in the configuration, but this is not necessarily the case, as apparently Spring can respond with the actual realm if it matches the allowed patterns (which can be also '*').

I think the reason, for making patterns explicit in Spring configuration is security, and the question is then if we want to make them explicit in the configuration as well? 
If not that we can just use `setAllowedOriginPatterns` for instead of `setAllowedOrigins` either for all configured origins (or these that include '*').

johngrimes:
This has been merged into `release/4.0.0`.

PR #317, opened 2021-06-28
Sum function and combine operator

johngrimes:
This change introduces a `sum` function for aggregating numeric data, and a combine operator for combining values from two collections into a single output collection.


fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/BooleanPath.java
```diff
@@ -77,4 +78,9 @@ public boolean isComparableTo(@Nonnull final Class<? extends Comparable> type) {
     return COMPARABLE_TYPES.contains(type);
   }
 
+  @Override
+  public boolean canBeCombinedWith(@Nonnull final FhirPath target) {
+    return super.canBeCombinedWith(target) || target instanceof BooleanLiteralPath;
+  }
+
```
piotrszul, line 85: This really feels somewhat sketchy (for all types) both because of the plethora of special cases and also because the semantics does not match 'mergeWith' in some case, e.g: two StringPaths can always be combined but the merge will fail if they are not of the same fhir type. 

Maybe each FhirPath should have a FhirDefinedType associated with it and then 'canBeCombined` could be in most cases just based on the equality (or equivalence of these types). It may need to be optional to account for UntypedResoucePath. 

A more complex solution would be to allow for implicit type conversions (e.g. Decimal + Integer) or ( URL + String).

Or maybe we should decouple the path hierarchy from the type hierarchy so that all type dependent operation are encapsulated in type classes... e.g. `BooleanType`, `StringType` etc. (including comparison, mateialisation etc),  and then we only have something like `class ElementPath<? extends Type> {  Type type}` which delegates all type related things to it's type object.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/parser/ParserTest.java
```diff
@@ -518,12 +515,22 @@ void testTranslateWithWhereAndTranslate() {
...
 
+  @Test
+  public void testCombineOperator() {
+    assertThatResultOf("name.family combine name.given")
+        .isElementPath(StringPath.class)
+        .selectResult()
+        .hasRows(spark, "responses/ParserTest/testCombineOperator.csv");
+  }
+
   @Test
```
piotrszul, line 526: Add a test for <element> combine <literal> and maybe for combining resources.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/operator/CombineOperatorTest.java
```diff
@@ -0,0 +1,243 @@
...
+            rowFromCoding(new Coding("http://snomed.info/sct", "373882004", null)))
+        .buildWithStructValue();
+    assertThat(result)
+        .hasExpression("valueCoding combine http://snomed.info/sct|373882004")
+        .isNotSingular()
+        .isElementPath(CodingPath.class)
+        .selectResult()
+        .hasRowsUnordered(expectedDataset);
+  }
+
```
piotrszul, line 222: Add test for combining resource paths.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/function/SumFunctionTest.java
```diff
@@ -0,0 +1,154 @@
...
+    final Dataset<Row> inputDataset = new DatasetBuilder(spark)
+        .withIdColumn()
+        .withEidColumn()
+        .withColumn(DataTypes.IntegerType)
+        .withRow("observation-1", makeEid(0), 3)
+        .withRow("observation-1", makeEid(1), 5)
+        .withRow("observation-1", makeEid(2), 7)
+        .withRow("observation-2", null, null)
+        .withRow("observation-3", makeEid(0), -1)
+        .build();
```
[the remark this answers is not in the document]
johngrimes: Fixed in 49bbfb6f8a0bbbc01e7f8549b355499cc26c4ce2

fhir-server/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java
```diff
@@ -0,0 +1,64 @@
...
+          Arrays.asList(operand.getIdColumn(), operand.getValueColumn()));
+      if (operand instanceof NonLiteralPath) {
+        final NonLiteralPath nonLiteralLeft = (NonLiteralPath) operand;
+        nonLiteralLeft.getThisColumn().ifPresent(columns::add);
+      }
+      input.getContext().getGroupingColumns().ifPresent(columns::addAll);
+      trimmed.add(operand.getDataset().select(columns.toArray(new Column[]{})));
+    }
+    final Dataset<Row> leftTrimmed = trimmed.get(0);
+    final Dataset<Row> rightTrimmed = trimmed.get(1);
```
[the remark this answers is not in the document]
johngrimes: Fixed in 6dcd28351b8a35fcba38e5dd38c7e8b6eaf791d8

fhir-server/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java
```diff
@@ -0,0 +1,64 @@
...
+
+    final Dataset<Row> dataset = leftTrimmed.union(rightTrimmed);
```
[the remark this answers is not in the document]
johngrimes: As far as I can tell both of these scenarios actually produce the correct result - I've added tests in eb0a0c5428d450d818940d453a4274e1c2358b1b and 5610e894a23b60c0b2e3143109c7006d9a5022d8.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java
```diff
@@ -0,0 +1,64 @@
...
+    final Optional<Column> thisColumn = left instanceof NonLiteralPath
+                                        ? ((NonLiteralPath) left).getThisColumn()
+                                        : Optional.empty();
+    return left
+        .mergeWith(right, dataset, expression, left.getIdColumn(),
+            Optional.empty(), left.getValueColumn(), false, thisColumn);
+  }
```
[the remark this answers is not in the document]
johngrimes: Add EID in acc90db0c2beff8bc48e2fb89ca8e674aaa530a2.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/operator/CombineOperatorTest.java
johngrimes: Add a ParserTest in 5610e894a23b60c0b2e3143109c7006d9a5022d8.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/parser/ParserTest.java
johngrimes: Fixed in eb0a0c5428d450d818940d453a4274e1c2358b1b and 5610e894a23b60c0b2e3143109c7006d9a5022d8.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/BooleanPath.java
johngrimes: I don't think that we want coercion as such - I don't think Decimal + Integer should be allowed, but I do think that URL + String should be allowed (as they are the same type in FHIRPath).

I wonder if `mergeWith` is being more strict than it needs to be?

I'll see what I can do to harmonise these concepts within the code.

johngrimes: I've made some changes in e990f3caf13759caea449210a2c861cd532c7315 which I hope go some way to addressing your concerns:

* Harmonise the combine/merge functionality used by iif and the combine operator into a single mechanism
* Broaden the set of valid cases to include the combination of paths which have the same schema representation
* Improve test coverage over some of the scenarios that you mentioned

PR #331, opened 2021-07-12
Prevent joins involving literal paths

johngrimes:
This change enhances the join logic to take advantage of scenarios where the expression on one side can be expressed purely using a column, as is in the case with literal paths.

Literal paths will now use a raw column (unaliased) which will be portable across any dataset.

A change was also required to the combine operator to preserve resource columns in cases where the this context is a resource. This was not identified as a problem before because expressions always included some other expression which carried the resource columns through (often a literal).


fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/LiteralPath.java
```diff
@@ -94,9 +92,8 @@ protected LiteralPath(@Nonnull final Dataset<Row> dataset, @Nonnull final Column
       @Nonnull final Type literalValue) {
     this.idColumn = idColumn;
     this.literalValue = literalValue;
-    final DatasetWithColumn datasetWithColumn = createColumn(dataset, buildValueColumn());
-    this.dataset = datasetWithColumn.getDataset();
-    this.valueColumn = datasetWithColumn.getColumn();
+    this.dataset = dataset;
+    this.valueColumn = buildValueColumn();
```
piotrszul, line 96: This looks like a small change but it may have some unexpected consequences.
For example if we do something like:
``` 1 combine 2 ``` I think results will be [1,1] instead of [1,2] because the value expression from the first literal will be become the value for the combination.
Maybe we should have a way to 'materialize' the literal (or any FhirPath) so that it all columns are in the dataset and all column references that we use (valueColumn) etc, are attribute references rather than expressions.

File: fhir-server/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java [hunk 2]
```diff
@@ -57,7 +57,15 @@ public FhirPath invoke(@Nonnull final OperatorInput input) {
...
+      // If there is a this column, we need to preserve it.
       nonLiteralLeft.getThisColumn().ifPresent(columns::add);
+      // If the this context is a resource, we need to preserve all element columns.
+      input.getContext().getThisContext().ifPresent(thisContext -> {
+        if (thisContext instanceof ResourcePath) {
+          final ResourcePath thisResource = (ResourcePath) thisContext;
+          columns.addAll(thisResource.getElementColumns());
+        }
+      });
     }
```
piotrszul, line 69: This looks a bit inelegant  with all the instance type checks. Maybe this should be a method on FhirPath? Either to trim or to produce the list of important columns for the path. 
Also I think some other path (UntypedResource ???) have some additional important columns (type I think).

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -272,32 +272,39 @@ private static boolean hasColumn(@Nonnull final Dataset<Row> dataset,
     Dataset<Row> dataset = left.getDataset();
     final List<FhirPath> joinTargets = fhirPaths.subList(1, fhirPaths.size());
 
+    boolean nonLiteralEncountered = left instanceof NonLiteralPath;
```
piotrszul, line 275: I think this works but is somewhat obscure. Maybe it would clearer if we just filter all NonLiterals up front and either apply the existing logic if there are any or return `fhirPaths.get(0)` if there are none.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -272,32 +272,39 @@ private static boolean hasColumn(@Nonnull final Dataset<Row> dataset,
...
-          // that they both originate from $this.
-          if (hasColumn(left.getDataset(), thisEidColumn)
-              && hasColumn(right.getDataset(), thisEidColumn)) {
-            leftColumns.add(thisEidColumn);
-            rightColumns.add(thisEidColumn);
+      if (right instanceof NonLiteralPath) {
+        if (nonLiteralEncountered) {
+          final List<Column> leftColumns = new ArrayList<>();
+          final List<Column> rightColumns = new ArrayList<>();
+          leftColumns.add(left.getIdColumn());
```
piotrszul, line 281: I am not sure if this is OK given that left may be a Literal and not be part of any of the previous joins.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
```diff
@@ -272,32 +272,39 @@ private static boolean hasColumn(@Nonnull final Dataset<Row> dataset,
...
+          // more rows than resources on either side of the join.
+          if (parserContext.getThisContext().isPresent()
+              && parserContext.getThisContext().get() instanceof NonLiteralPath) {
+            final NonLiteralPath nonLiteralThis = (NonLiteralPath) parserContext.getThisContext()
+                .get();
+            if (nonLiteralThis.getEidColumn().isPresent()) {
+              final Column thisEidColumn = nonLiteralThis.getEidColumn().get();
+              // Only use the element ID for the join if both datasets have a $this column, which means
+              // that they both originate from $this.
+              if (hasColumn(left.getDataset(), thisEidColumn)
```
piotrszul, line 295: Same as above - if left is literal we will never use eid for joins even if we should . 
But this is actually wrong even without this change (I think). The test should be not whether left has eid, but if the current join result (dataset) has eid.

piotrszul (commented):
I think I have spotted a few small issue but it's likely that that they are only relevant for some 'exotic' cases.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/LiteralPath.java
johngrimes: Fixed in f6fe23340eb56c034508b9d5fe5e7bc87a1bc9aa.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: Fixed in 59295803b97509c48ae9620dc3f13809c425b80a.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: Fixed in 59295803b97509c48ae9620dc3f13809c425b80a.

fhir-server/src/main/java/au/csiro/pathling/QueryHelpers.java
johngrimes: Fixed in 59295803b97509c48ae9620dc3f13809c425b80a.

PR #365, opened 2021-08-23
Implement extract operation

johngrimes:
This change will introduce a new operation called `extract`. This operation is designed for transforming FHIR data into a flattened form, for use within other tools, such as statistical and machine learning models.

Still to do:

- [ ] Look into Spring Boot caching
- [ ] Allow for extract of Coding values
- [ ] Add support for async response
- [ ] Add tests

### DIFF
--- a/fhir-server/pom.xml
+++ b/fhir-server/pom.xml
@@ -145,6 +145,12 @@
       <version>${pathling.antlrVersion}</version>
     </dependency>
 
+    <!-- AWS SDK -->
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-s3</artifactId>
+    </dependency>
+
     <!-- Error monitoring -->
     <dependency>
       <groupId>io.sentry</groupId>
--- a/fhir-server/src/main/java/au/csiro/pathling/Configuration.java
+++ b/fhir-server/src/main/java/au/csiro/pathling/Configuration.java
@@ -176,6 +176,12 @@ public class Configuration {
       @ToString.Exclude
       private String secretAccessKey;
 
+      /**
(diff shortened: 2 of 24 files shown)


johngrimes:
I think I have finally cracked the problem of how to join the column expressions, taking into account element identity (thanks to an algorithm from the brain of @piotrszul).

I just need to implement a filtering solution now, which I think is going to be something very similar to what is used within the SearchExecutor (get a list of filtered IDs, and use a left semi-join).

johngrimes:
@piotrszul I think this is ready for you to have a look at - thanks.

johngrimes:
There are two problems remaining with this implementation:

1. Async and authentication don't work together, specifically in the retrieval of the async result via the `$job` operation. I think this is related to AOP tripping over itself. I need to add an integration test that reproduces this.
2. In order for this to work with Power BI (and perhaps other real-world integrations that have a problem with talking to multiple endpoints/protocols), we need to proxy the retrieval of the result. This adds complexity but I think it is necessary. I'm proposing to create a new `$result` operation that takes an ID. This operation will look up an in-memory structure which will contain the actual URL and the MIME type and stream it back to the client.

johngrimes:
@piotrszul I've got async and authorisation working together, but now these two tests are failing for reasons that I can't fathom. It seems to be working as intended when tested manually.

- `SecurityEnabledOperationsTest.testForbiddenIfAggregateWithoutAuthority `
- `SecurityEnabledOperationsTest.testForbiddenIfImportWithoutAuthority`

It might be related to the fact that I needed to separate out the methods that were targeted by AOP in order to get it all working.

johngrimes:
I'm going to merge this into `release/5.0.0` now, I think this branch has lived long enough :slightly_smiling_face:

PR #398, opened 2021-11-04
GA4GH passports and visas

johngrimes:
This is the branch that we're going to use to work on this integration, and we can use this pull request to capture the related discussion.


johngrimes:
@ametke @andrewpatto I think I have got this all working with v1 passports, including all the construction and enforcement of scoped queries.

You can now present a passport instead of a regular token, and the implementation will translate the manifest into a set of authorities and scopes.

Still to do is testing, as well as coming up with some data that we can use to demonstrate this functionality.

There are also some optimisations that can still be make to the filter expressions themselves, the way it is currently set up is quite inefficient from a query performance perspective.

Once this is all built, I will deploy it to the demo instance.

johngrimes:
I have deployed these changes to https://nagim.pathling.app/fhir.

It is configured to accept passports issued by https://broker.nagim.dev, and visas issued by https://didact-patto.dev.umccr.org or https://didact-nagim.dev.umccr.org.

johngrimes:
The passports functionality is activated using the `ga4gh` Spring profile.

You would run it using the following environment variable (or system property):

```
spring.profiles.active=core,server,ga4gh
```

Issue #341, opened 2021-07-21
Date/time arithmetic


johngrimes:
This change builds upon the work in #340 to add support for the [Date/Time Arithmetic](https://hl7.org/fhirpath/#datetime-arithmetic) described in the FHIRPath specification.

PR #509, opened 2022-05-01
Quantities and date arithmetic

### LINKED ISSUE #340: Quantity type
This change implements support for the Quantity type, including literals, support for equality, equivalence and comparison operators, and support for its use within grouping.

johngrimes:
Resolves #340 and #341.


johngrimes:
I don't think we need to encode scale against the canonicalized Quantity value, as we only have to preserve the scale of original field.

johngrimes:
The core functionality is now working, here is an example query: https://go.pathling.app/extract?subjectResource=Observation&columns%5B0%5D%5Blabel%5D=Observation%20ID&columns%5B0%5D%5Bexpression%5D=id&columns%5B1%5D%5Blabel%5D=Observation%20type&columns%5B1%5D%5Bexpression%5D=code.coding.display&columns%5B2%5D%5Blabel%5D=Value&columns%5B2%5D%5Bexpression%5D=valueQuantity.value&columns%5B3%5D%5Blabel%5D=Unit&columns%5B3%5D%5Bexpression%5D=valueQuantity.code&columns%5B4%5D%5Blabel%5D=Less%20than%206.4%20mmol%2FL&columns%5B4%5D%5Bexpression%5D=valueQuantity%20%3C%206.4%20%27mmol%2FL%27

johngrimes:
The tasks outstanding are:

- Ensure test coverage
- Add `until` function
- Add `toQuantity` function
- Update documentation

johngrimes:
We've taken the approach of pre-baking a canonicalized version of each UCUM Quantity using the encoders. This is because it is expensive to calculate this on the fly for large datasets, and UCUM rarely changes.

We've added a `_value_canonicalized` and `_code_canonicalized` field to the encoding of Quantity. This has the potential to benefit users of the encoders, as well.

johngrimes:
These new features will require an upstream change before we can merge: FHIR/Ucum-java#28

johngrimes:
We still have a couple of outstanding TODOs from a previous code review:

- [ ] `au.csiro.pathling.encoders.SerializerBuilderProcessor#proceedCompositeChildren` Maybe create specialized UCUM functions returning spark.Decimal and UTF8String
- [ ] `au.csiro.pathling.encoders.SerializerBuilderProcessor#proceedCompositeChildren` Maybe move to overrideCompositeExpression() providing it with a callback to generate default fields on request (a lazy call to super)

johngrimes:
@piotrszul Would you mind taking a look at the outstanding TODOs in the previous comment, and giving this PR a final check?

encoders/src/main/scala/au/csiro/pathling/encoders/SchemaConverter.scala
```diff
@@ -55,7 +56,14 @@ private[encoders] class SchemaConverterProcessor(override val fhirContext: FhirC
...
-    StructType(fields ++ createFidField() ++ createExtensionField(definition))
+    val updatedFields = definition.getImplementingClass match {
+      case cls if classOf[Quantity].isAssignableFrom(cls) => fields ++ Seq(
+        StructField("_value_canonicalized", DecimalCustomCoder.decimalType),
+        StructField("_code_canonicalized", DataTypes.StringType)
+      )
+      case _ => fields
+    }
+    StructType(updatedFields ++ createFidField() ++ createExtensionField(definition))
   }
```
piotrszul, line 68: Short term maybe it would be better to refactor the quality fields creation (conditional) to a method similar to `createExtensionsFields` and write something lie:
`StructType(fields ++ createQuantityFields(definition) ++ createFidField() ++ createExtensionField(definition))`

Longer term this perhaps can be designed in such a way that all aspects of creation of synthetic (i.e. schema, ser, deser) is encapsulated in a single class, so that changes to that do not need to be tracked across three differenc classes (i.e. SchemaConverter, SerializderBuilder and DeserializerBuider).

encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
```diff
@@ -107,19 +108,42 @@ private[encoders] class SerializerBuilderProcessor(expression: Expression,
...
   override def proceedCompositeChildren(value: CompositeCtx[Expression, (String, Expression)]): Seq[(String, Expression)] = {
-    dataTypeMappings.overrideCompositeExpression(expression, value.compositeDefinition)
-      .getOrElse(super.proceedCompositeChildren(value))
+
+    value.compositeDefinition.getImplementingClass match {
+      case cls if classOf[Quantity].isAssignableFrom(cls) =>
+        val valueExp = Invoke(expression, "getValue", ObjectType(classOf[java.math.BigDecimal]))
+        val codeExp = Invoke(expression, "getCode", ObjectType(classOf[java.lang.String]))
+        // TODO: Maybe create specialized UCUM functions returning spark.Decimal and UTF8String
+        // TODO: Maybe move to overrideCompositeExpression() providing it with a callback to generate default fields on request (a lazy call to super)
```
piotrszul, line 117: Short term maybe just refactor our the creation of the extra serializers, 
e.g. .
```getOrElse(super.proceedCompositeChildren(value) ++  createSynhteticSerialzers(value))```

Longer term consider together with schema (see the comment in schema converter)
This can be done in `overrideCompositeExpression` it's signature is change to:
```
def overrideCompositeExpression(inputObject: Expression,
    definition: BaseRuntimeElementCompositeDefinition[_], 
    defaultFields:  => Seq[ExpressionWithName] // this is a lazy argument only evaluated it actually used
     ):Option[Seq[ExpressionWithName]]
```

File: encoders/src/test/java/au/csiro/pathling/encoders/LightweightFhirEncodersTest.java [hunk 3]
```diff
@@ -205,4 +208,55 @@ public void testEncodesExtensions() {
...
+        encoder);
+    final InternalRow serializedRow = resolvedEncoder.createSerializer().apply(observation);
+
+    final ExpressionEncoder<Row> rowEncoder = EncoderUtils.defaultResolveAndBind(
+        RowEncoder.apply(encoder.schema()));
+    final Row observationRow = rowEncoder.createDeserializer().apply(serializedRow);
+
+    final Row quantityRow = observationRow.getStruct(observationRow.fieldIndex("valueQuantity"));
+    assertQuantity(quantityRow, "76000.000000", "g");
+  }
```
piotrszul, line 227: Maybe it would be better to test agains the original quantity making sure that both the canocalized and original fields are encoded/decoded correctly.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/Temporal.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+import org.apache.spark.sql.Column;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.functions;
+import org.hl7.fhir.r4.model.Enumerations.FHIRDefinedType;
+
+public interface Temporal {
```
piotrszul, line 21: Add javadocs.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/UntilFunction.java
```diff
@@ -0,0 +1,66 @@
...
+import au.csiro.pathling.fhirpath.literal.DateTimeLiteralPath;
+import au.csiro.pathling.fhirpath.literal.StringLiteralPath;
+import au.csiro.pathling.sql.dates.TemporalDifferenceFunction;
+import javax.annotation.Nonnull;
+import org.apache.spark.sql.Column;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.hl7.fhir.r4.model.Enumerations.FHIRDefinedType;
+
+public class UntilFunction implements NamedFunction {
```
piotrszul, line 29: Add Javadocs

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/UntilFunction.java
```diff
@@ -0,0 +1,66 @@
...
+
+    final Dataset<Row> dataset = join(input.getContext(), fromArgument, toArgument,
+        JoinType.LEFT_OUTER);
+    final Column valueColumn = callUDF(TemporalDifferenceFunction.FUNCTION_NAME,
+        fromArgument.getValueColumn(), toArgument.getValueColumn(),
+        calendarDurationArgument.getValueColumn());
+    final String expression = NamedFunction.expressionFromInput(input, NAME);
+
+    return ElementPath.build(expression, dataset, fromArgument.getIdColumn(),
+        fromArgument.getEidColumn(), valueColumn, fromArgument.isSingular(),
```
piotrszul, line 62: I think that the singularity of the results may be different than `fromArgument.isSingular()` specifically when `toArgument` is not singular.  I am not sure how we would actually handle such case, so perhaps we should require the toArgument to be singular.  Or maybe it should somehow be related to  eid. I just think that in general it's hard to predict what the join with toArgument yields both with in terms of singularity with regard to id and eid.

fhir-server/src/test/java/au/csiro/pathling/test/UnitTestDependencies.java
```diff
@@ -38,7 +61,18 @@ class UnitTestDependencies {
...
+        new DateSubtractDurationFunction(),
+        new DateTimeEqualsFunction(), new DateTimeGreaterThanFunction(),
+        new DateTimeGreaterThanOrEqualToFunction(), new DateTimeLessThanFunction(),
+        new DateTimeLessThanOrEqualToFunction(), new TimeEqualsFunction(),
+        new TimeGreaterThanFunction(), new TimeGreaterThanOrEqualToFunction(),
+        new TimeLessThanFunction(), new TimeLessThanOrEqualToFunction());
+    final List<SqlFunction3> sqlFunction3 = List.of(new TemporalDifferenceFunction());
+    return Spark.build(configuration, environment, sparkListener, sqlFunction1, sqlFunction2,
+        sqlFunction3);
   }
```
piotrszul, line 76: All these functions can probably be found with 'ListableBeanFactory` and I believe there should be a way to get access to it in this method (normally this would be with the 'BeanFactoryAware').

fhir-server/src/main/java/au/csiro/pathling/sql/udf/SqlFunction.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright © 2018-2022, Commonwealth Scientific and Industrial Research
+ * Organisation (CSIRO) ABN 41 687 119 230. Licensed under the CSIRO Open Source
+ * Software Licence Agreement.
+ */
+
+package au.csiro.pathling.sql.udf;
+
+import org.apache.spark.sql.types.DataType;
+
+public interface SqlFunction {
```
piotrszul, line 11: Add Javadocs

fhir-server/src/main/java/au/csiro/pathling/sql/udf/SqlFunction1.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © 2018-2022, Commonwealth Scientific and Industrial Research
+ * Organisation (CSIRO) ABN 41 687 119 230. Licensed under the CSIRO Open Source
+ * Software Licence Agreement.
+ */
+
+package au.csiro.pathling.sql.udf;
```
piotrszul, line 7: Add Javadoc ???

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DateTimePath.java
```diff
@@ -80,41 +81,53 @@ public static Optional<BaseDateTimeType> valueFromRow(@Nonnull final Row row,
...
+          functionName = DateTimeGreaterThanFunction.FUNCTION_NAME;
+          break;
+        case GREATER_THAN_OR_EQUAL_TO:
+          functionName = DateTimeGreaterThanOrEqualToFunction.FUNCTION_NAME;
+          break;
+        default:
+          throw new AssertionError("Unsupported operation: " + operation);
+      }
+      return callUDF(functionName, source.getValueColumn(), target.getValueColumn());
+    };
```
piotrszul, line 130: Perhaps delegate to Comparator (DataTimeComparator) see: `spike/340`

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DecimalPath.java
```diff
@@ -146,7 +151,7 @@ public static Function<Numeric, NonLiteralPath> buildMathOperation(@Nonnull fina
           valueColumn = valueColumn.cast(getDecimalType());
           return ElementPath
               .build(expression, dataset, idColumn, eidColumn, valueColumn, true, Optional.empty(),
-                  thisColumn, fhirType);
+                  thisColumn, source.getFhirType());
         case MODULUS:
           valueColumn = valueColumn.cast(DataTypes.LongType);
           return ElementPath
```
piotrszul, line 157: Perhaps can be delegated to a counterpart of Comparator for arithmetic operations.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/IntegerPath.java
```diff
@@ -157,7 +165,7 @@ public static Function<Numeric, NonLiteralPath> buildMathOperation(@Nonnull fina
           }
           return ElementPath
               .build(expression, dataset, idColumn, eidColumn, valueColumn, true, Optional.empty(),
-                  thisColumn, fhirType);
+                  thisColumn, source.getFhirType());
         case DIVISION:
           final Column numerator = source.getValueColumn().cast(DecimalPath.getDecimalType());
           valueColumn = operation.getSparkFunction().apply(numerator, targetValueColumn);
```
piotrszul, line 171: Perhaps can be delegated to a counterpart of Comparator for arithmetic operations.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/QuantityPath.java
```diff
@@ -30,4 +46,102 @@ protected QuantityPath(@Nonnull final String expression, @Nonnull final Dataset<
...
+          QuantityEncoding.CANONICALIZED_CODE_COLUMN);
+      final Column targetCode = comparableTarget.getField(
+          QuantityEncoding.CANONICALIZED_CODE_COLUMN);
+      final Column sourceValue = comparableSource.getField(
+          QuantityEncoding.CANONICALIZED_VALUE_COLUMN);
+      final Column targetValue = comparableTarget.getField(
+          QuantityEncoding.CANONICALIZED_VALUE_COLUMN);
+      final Column compareValues = operation.getSparkFunction().apply(sourceValue, targetValue);
+      return when(sourceCode.equalTo(targetCode), compareValues).otherwise(null);
+    };
```
piotrszul, line 71: Perhaps can be delegate to `QuantityComparator` (see: spike/340)

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/QuantityPath.java
```diff
@@ -30,4 +46,102 @@ protected QuantityPath(@Nonnull final String expression, @Nonnull final Dataset<
...
+        case MULTIPLICATION:
+        case DIVISION:
+          return ElementPath
+              .build(expression, dataset, idColumn, eidColumn, resultQuantityColumn, true,
+                  Optional.empty(),
+                  thisColumn, fhirType);
+        default:
+          throw new AssertionError("Unsupported math operation encountered: " + operation);
+      }
+    };
```
piotrszul, line 144: Perhaps can be delegated to a counterpart of Comparator for arithmetic operations.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/encoding/QuantityEncoding.java
```diff
@@ -0,0 +1,98 @@
...
+            DecimalCustomCoder.precision(), DecimalCustomCoder.scale()), true, metadata);
+    final StructField canonicalizedCode = new StructField(CANONICALIZED_CODE_COLUMN,
+        DataTypes.StringType, true, metadata);
+    final StructField fid = new StructField("_fid", DataTypes.IntegerType, true,
+        metadata);
+    return new StructType(
+        new StructField[]{id, value, valueScale, comparator, unit, system, code, canonicalizedValue,
+            canonicalizedCode, fid});
+  }
+
```
piotrszul, line 97: This (and other encoding like this, e.g. for SimpleCoding) should be placed closer to the `encoders` module.
Minimally the encoders should provide an interface that encapsulates encode/decode/datatype operations (possibly with the current implementation).
Ideally we should be able to use the builders/converters in encoders to provide the implementation.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/QuantityLiteralPath.java
```diff
@@ -6,92 +6,225 @@
...
-        final BigDecimal value = DecimalLiteralPath.fromString(fhirPath, context)
-            .getLiteralValue().getValue();
-        quantity.setValue(value);
-      } catch (final NumberFormatException ex) {
-        throw new IllegalArgumentException("Quantity literal has invalid format: " + fhirPath);
-      }
+      throw new IllegalArgumentException("Quantity literal has invalid value: " + value);
     }
+    return decimalValue;
+  }
```
piotrszul, line 132: I think both of the parsing methods above could be refactored out to `QuantityLiteral` in the same way we have `CodingLiteral`

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/QuantityLiteralPath.java
```diff
@@ -6,92 +6,225 @@
...
+        lit(value).as("value"),
+        lit(value.scale()).as("value_scale"),
+        lit(comparator.map(QuantityComparator::toCode).orElse(null)).as("comparator"),
+        lit(quantity.getUnit()).as("unit"),
+        lit(quantity.getSystem()).as("system"),
+        lit(quantity.getCode()).as("code"),
+        lit(canonicalizedValue).as(QuantityEncoding.CANONICALIZED_VALUE_COLUMN),
+        lit(canonicalizedCode).as(QuantityEncoding.CANONICALIZED_CODE_COLUMN),
+        lit(null).as("_fid"));
+  }
```
piotrszul, line 191: Another place, which implicitly depends on the internals of encoders (but so does CodingLiteralPath).
Perhaps it could be moved (at least partially to QuantityEncoder - for the creation of the literal struct).
Or in it's entirety to `QuantityLiteral` if we introduce one (see the comment above)

fhir-server/src/main/java/au/csiro/pathling/spark/Spark.java
```diff
@@ -43,14 +45,20 @@ public class Spark {
...
   @ConditionalOnMissingBean
   @Nonnull
   public static SparkSession build(@Nonnull final Configuration configuration,
       @Nonnull final Environment environment,
-      @Nonnull final Optional<SparkListener> sparkListener) {
+      @Nonnull final Optional<SparkListener> sparkListener,
+      @Nonnull final List<SqlFunction1> sqlFunction1,
+      @Nonnull final List<SqlFunction2> sqlFunction2,
+      @Nonnull final List<SqlFunction3> sqlFunction3) {
     log.debug("Creating Spark session");
```
piotrszul, line 62: Does Spring automatically create these lists from all the compatible components?

fhir-server/src/test/java/au/csiro/pathling/test/helpers/SparkHelpers.java
```diff
@@ -98,6 +104,31 @@ public static StructType referenceStructType() {
...
+        DataTypes.createDecimalType(
+            DecimalCustomCoder.precision(), DecimalCustomCoder.scale()), true, metadata);
+    final StructField fid = new StructField("_fid", DataTypes.IntegerType, true,
+        metadata);
+    return new StructType(
+        new StructField[]{id, value, valueScale, comparator, unit, system, code, canonicalizedValue,
+            canonicalizedCode, fid});
+  }
+
   @Nonnull
```
piotrszul, line 132: `Seems to be the same as QuantityEncoding.dataType()`

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
```diff
@@ -0,0 +1,109 @@
...
+        .intValue();
+    final Integer temporalUnit = TemporalArithmeticFunction.CALENDAR_DURATION_TO_UCUM.get(
+        calendarDuration.getCode());
+    checkUserInput(temporalUnit != null,
+        "Unsupported calendar duration unit: " + calendarDuration.getCode());
+
+    @SuppressWarnings("unchecked")
+    final IntermediateType result = (IntermediateType) temporal.copy();
+    result.add(temporalUnit, subtract
+                             ? amountToAdd * -1
```
piotrszul, line 80: `amountToAdd * -1` is the same as `- amountToAdd`

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
```diff
@@ -0,0 +1,109 @@
...
+      throw new IllegalArgumentException("Calendar duration must have a system of "
+          + QuantityLiteralPath.FHIRPATH_CALENDAR_DURATION_URI);
+    }
+    final int amountToAdd = calendarDuration.getValue()
+        .setScale(0, RoundingMode.HALF_UP)
+        .intValue();
+    final Integer temporalUnit = TemporalArithmeticFunction.CALENDAR_DURATION_TO_UCUM.get(
+        calendarDuration.getCode());
+    checkUserInput(temporalUnit != null,
+        "Unsupported calendar duration unit: " + calendarDuration.getCode());
```
piotrszul, line 75: Maybe this conversion could be refactored out to some kind 'UcumUtils`.

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
```diff
@@ -0,0 +1,109 @@
...
+import javax.annotation.Nullable;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.hl7.fhir.r4.model.BaseDateTimeType;
+import org.hl7.fhir.r4.model.Quantity;
+
+public abstract class TemporalArithmeticFunction<IntermediateType extends BaseDateTimeType> implements
+    SqlFunction2<String, Row, String> {
+
```
piotrszul, line 30: I think that the oracle convention for the names of type parameters of generic types is a single upper case letter.
This is probably a matter of preference but I do find the use of 'IntermediateType` somewhat confusing, when I read the code as it just does not register as a type parameter but rather a type itself.

piotrszul (commented):
Hi, I finished review the code - an epic piece of work!
I did not find any obvious issues. There are a few places where javadoc is missing.
Most of my comments are just ideas of improvements.
I have checked in some of my ideas on to `spike/340` - it mostly deals with two aspects :

- the interface between FhirPaths and comparison/arithmetic operations. I think that it can be more explicit and unified so that every Comparable path works in the same way and the differences are encapsulated in the corresponding `Comparator`.
- the alternative approach for defining (and registration of) our UDFs.

They are somewhat independent and considered separately (happy to have a chat about it).

I am also considering if the testing of the UDFs should not be more focused on unit testing the UDF operations themselves rather that the FhirPaths that use them. Or at least it would be probably easier to write more through tests cases for the UDFs (regardless if they are beans or functions) than for FhirPaths. And also I assume that in the future we my want/need to expose these UDFs in the python API (so perhaps they should be tested independently from FhirPaths).

encoders/src/main/scala/au/csiro/pathling/encoders/SchemaConverter.scala
piotrszul: Refctored.

Long term solution captured in #842 [Review and harmonize encoding customisations]

encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
piotrszul: Refactored.

Long term solution captured in https://github.com/aehrc/pathling/issues/842 [Review and harmonize encoding customisations]

encoders/src/test/java/au/csiro/pathling/encoders/LightweightFhirEncodersTest.java
piotrszul: This is actually canonicalisation test, so the assertion is appropriate.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DateTimePath.java
piotrszul: Refactored out to the comparator.

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
piotrszul: Fixed

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
piotrszul: Type name changed to T.

fhir-server/src/test/java/au/csiro/pathling/test/helpers/SparkHelpers.java
piotrszul: Fixed.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/QuantityPath.java
piotrszul: Fixed.

fhir-server/src/main/java/au/csiro/pathling/spark/Spark.java
piotrszul: Yes it does.

fhir-server/src/test/java/au/csiro/pathling/test/UnitTestDependencies.java
piotrszul: Fixed.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/QuantityLiteralPath.java
piotrszul: Moved to QuantityEncoder.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/Temporal.java
piotrszul: Added.

fhir-server/src/main/java/au/csiro/pathling/sql/udf/SqlFunction.java
piotrszul: Added java docs.

fhir-server/src/main/java/au/csiro/pathling/sql/udf/SqlFunction1.java
piotrszul: Added Java docs.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/function/UntilFunction.java
piotrszul: Made 'until' to require the singular object and argument (in the same way) as the arithmetic operators.

fhir-server/src/main/java/au/csiro/pathling/fhirpath/encoding/QuantityEncoding.java
piotrszul: Captured as #868

fhir-server/src/main/java/au/csiro/pathling/sql/dates/TemporalArithmeticFunction.java
piotrszul: Refactored to CalendarDurationUtils

fhir-server/src/main/java/au/csiro/pathling/fhirpath/literal/QuantityLiteralPath.java
piotrszul: Parsing of calendar duration refactored to CalendarDuationUtils. Ucum quantities parsing can be refactored later is need for python interface (see CodingLiteral).

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/DecimalPath.java
piotrszul: Ignored (not obvious how that could be done).

fhir-server/src/main/java/au/csiro/pathling/fhirpath/element/IntegerPath.java
piotrszul: Ignored (not obvious how that could be done).

johngrimes:
I think there is some missing test coverage relating to the use of Integer or Decimal types with a Quantity and a math operator. Currently an expression such as `1 * 1 'm'` causes an error. It passes validation as both paths are now Numeric.

I will add some more tests to cover this, and make the appropriate adjustments to the implementation.

johngrimes:
In cases where one operand is a Quantity and the other is an Integer or Decimal, I think the correct solution to this is to convert the Integer or Decimal to a Quantity with a unit of `'1'`, 

In the [Operators > Math](https://hl7.org/fhirpath/#math-2) section, the spec says:

> Both operands must be of the same type, or of compatible types according to the rules for implicit conversion.

The [conversion table](https://hl7.org/fhirpath/#conversion) says that Integer and Decimal types can be implicitly converted to Quantity.

johngrimes:
After a brief trip down that rabbit hole, I worked out that implementing that part of the specification actually provides very little value.

I have resolved this problem by adding a new test: `MathOperatorValidationTest#operandsAreNotComparable`.

I have also added additional validation to `MathOperator` to check that the operands are comparable.

johngrimes:
@piotrszul There seem to be two classes, `FlexiDecimal` and `FlexDecimal`.

`FlexDecimal` seems to only be used within the decimal benchmark, I am going to assume that this is a mistake and that we should be using `FlexiDecimal` everywhere...

johngrimes:
Or were these two different implementations that you were comparing?

Can we retire one of them now?

piotrszul:
> After a brief trip down that rabbit hole, I worked out that implementing that part of the specification actually provides very little value.
> 
> I have resolved this problem by adding a new test: `MathOperatorValidationTest#operandsAreNotComparable`.
> 
> I have also added additional validation to `MathOperator` to check that the operands are comparable.

Perhaps this can be addressed (the implicit conversions) when (if ever) we implement the explicit conversion functions (e.g.: toQuantity()). Because then it would be just the matter of 'implicitly' applying this function in certain situations.

johngrimes:
> > After a brief trip down that rabbit hole, I worked out that implementing that part of the specification actually provides very little value.
> > I have resolved this problem by adding a new test: `MathOperatorValidationTest#operandsAreNotComparable`.
> > I have also added additional validation to `MathOperator` to check that the operands are comparable.
> 
> Perhaps this can be addressed (the implicit conversions) when (if ever) we implement the explicit conversion functions (e.g.: toQuantity()). Because then it would be just the matter of 'implicitly' applying this function in certain situations.

Yes, I have a good idea of how it could be done - I'm just not sure that we have a solid use case for it at the moment.

piotrszul:
> Or were these two different implementations that you were comparing?
> 
> Can we retire one of them now?

Yes, we can retire the string based implementation (FlexDecimal). As for the benchmark I would prefer to retain it (for decimal and FlexiDecimal) but just do not include it in the build. It may still be useful to run it locally to possibly further optimize FlexiDeciamal performance. The easies way to achieve it is to rename the class so that it does not fit the patters from `PathlingBenchmarkRunner` (.include("\\.[^.]+Benchmark\\.[^.]+$")).
Or maybe move such local benchmark to a different package and adjust the patter above.

johngrimes:
Ok, I have kept the DecimalBenchmark in but disabled, and with FlexDecimal removed.

PR #632, opened 2022-06-17
Elimination of invalid groupings based on common ancestor in aggregation.

### LINKED ISSUE #542: Aggregation not eliminating invalid groupings based on common ancestor
This change relates to removing invalid combinations of grouping values when 
aggregations are executed over paths that share a common ancestor element.

This is show by the following example. Given a set of Observation resources:

| id  | name.given | name.family |
|-----|------------|-------------|
| 1   | Benjamin   | Franklin    |
| 1   | Silence    | Dogood      |
| 2   | Isaac      | Asimov      |
| 2   | Paul       | French      |

And the following query:

```json
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "aggregation",
      "valueString": "count()"
    },
    {
      "name": "grouping",
      "valueString": "name.given"
    },
    {
      "name": "grouping",
      "valueString": "name.family"
    }
  ]
}
```

We should get the following result:

| Given name | Family name | Number of patients |
|------------|-------------|--------------------|
| Benjamin   | Franklin    | 1                  |
| Silence    | Dogood      | 1                  |
| Isaac      | Asimov      | 1                  |
| Paul       | French      | 1                  |

Not this:

| Given name | Family name | Number of patients |
|------------|-------------|--------------------|
| Benjamin   | Franklin    | 1                  |
| Benjamin   | Dogood      | 1                  |
| Silence    | Franklin    | 1                  |
| Silence    | Dogood      | 1                  |
| Isaac      | Asimov      | 1                  |
| Isaac      | French      | 1                  |
| Paul       | Asimov      | 1                  |
| Paul       | Frence      | 1                  |

This is not technically a bug, but it is a refinement of the way that the Aggregate operation works, and it brings it in line with the Extract column joining logic.

piotrszul:
Resolves #542.

### DIFF
--- a/fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
+++ b/fhir-server/src/main/java/au/csiro/pathling/QueryExecutor.java
@@ -6,6 +6,15 @@
 
 package au.csiro.pathling;
 
+import static au.csiro.pathling.QueryHelpers.join;
+import static au.csiro.pathling.utilities.Preconditions.check;
+import static au.csiro.pathling.utilities.Preconditions.checkArgument;
+import static au.csiro.pathling.utilities.Preconditions.checkNotNull;
+import static au.csiro.pathling.utilities.Preconditions.checkUserInput;
+import static au.csiro.pathling.utilities.Strings.randomAlias;
+import static org.apache.spark.sql.functions.coalesce;
+import static org.apache.spark.sql.functions.col;
+
 import au.csiro.pathling.QueryHelpers.DatasetWithColumn;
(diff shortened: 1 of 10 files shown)


johngrimes (changes requested):
This looks good.

At this point, I think we should continue on and define some more tests that target any parts of the implementation that we think might not be consistent with this method of combining groupings.

Then we can make sure that if there are any further changes to the query logic, they get done now as part of this change.

johngrimes:
The aggregate operation has now been deprecated in favour of SQL on FHIR query functionality.

Issue #603, opened 2022-06-13
Python: subsumes / subsumedBy functions


johngrimes:
This change adds a function to the Python library that corresponds to the [subsumes](https://pathling.csiro.au/docs/fhirpath/functions#subsumes) and [subsumedBy](https://pathling.csiro.au/docs/fhirpath/functions#subsumedby) FHIRPath functions.

This would allow Python users to add a column to a Spark dataframe that represents the result of a subsumption test on an input column.

This is the intended method documentation, this would be an instance method on `PathlingContext`:

```
subsumes(df: DataFrame, inputColumn: Column, targetCoding: str, outputColumnName: str) → DataFrame
subsumedBy(df: DataFrame, inputColumn: Column, targetCoding: str, outputColumnName: str) → DataFrame
```

Takes a dataframe with a Coding column as input. A new column is created which contains a Boolean value, indicating whether the input Coding subsumes/is subsumed by the specified Coding. The target Coding is described using a [Coding literal](https://pathling.csiro.au/docs/fhirpath/data-types#coding).

### Parameters

**df** - a DataFrame containing the input data
**codingColumn** - a Column containing a struct representation of a Coding
**targetCoding** - the Coding that is the target of the test
**outputColumnName** - the name of the result column

### Returns

A new dataframe with an additional column containing the result of the operation.

This does not yet have the ability to provide the target Coding(s) via a column - this will be the subject of a future enhancement.

Issue #602, opened 2022-06-13
Python: translate function


johngrimes:
This change adds a function to the Python library that corresponds to the [translate](https://pathling.csiro.au/docs/fhirpath/functions#translate) FHIRPath function.

This would allow Python users to add a column to a Spark dataframe that represents the translation of a coded column using a nominated FHIR ConceptMap.

This is the intended method documentation, this would be an instance method on `PathlingContext`:

```
translate(df: DataFrame, codingColumn: Column, conceptMapUri: str, reverse: Optional[bool] = False, equivalence: Optional[str] = "equivalent", outputColumnName: str) → DataFrame
```

Takes a dataframe with a Coding column as input. A new column is created which contains a Coding value and contains translation targets from the specified FHIR ConceptMap. There may be more than one target concept for each input concept.

### Parameters

**df** - a DataFrame containing the input data
**codingColumn** - a Column containing a struct representation of a Coding
**conceptMapUri** - an identifier for a FHIR ConceptMap
**reverse** - the direction to traverse the map - `false` results in "source to target" mappings, while `true` results in "target to source"
**equivalence** - a comma-delimited set of values from the [ConceptMapEquivalence](https://www.hl7.org/fhir/R4/valueset-concept-map-equivalence.html) ValueSet
**outputColumnName** - the name of the result column

### Returns

A new dataframe with an additional column containing the result of the operation.

johngrimes:
We have a prototype implementation of this on the `spike/tlg_python` branch.

PR #653, opened 2022-06-25
Implement member_of, translate and subsumes functions

### LINKED ISSUE #601: Python: memberOf function
This change adds a function to the Python library that corresponds to the [memberOf](https://pathling.csiro.au/docs/fhirpath/functions#memberof) FHIRPath function.

This would allow Python users to add a column to a Spark dataframe that represents the membership of a coded column within a nominated FHIR ValueSet.

This is the intended method documentation, this would be an instance method on `PathlingContext`:

```
memberOf(df: DataFrame, codingColumn: Column, valueSetUrl: str, outputColumnName: str) → DataFrame
```

Takes a dataframe with a Coding column as input. A new column is created which contains a Boolean value, indicating whether the input Coding is a member of the specified FHIR ValueSet.

### Parameters

**df** - a DataFrame containing the input data
**codingColumn** - a Column containing a struct representation of a Coding
**valueSetUrl** - an identifier for a FHIR ValueSet
**outputColumnName** - the name of the result column

### Returns

A new dataframe with an additional column containing the result of the operation.

johngrimes:
This change adds `member_of`, `translate` and `subsumes` functions to the Python API, allowing users to make use of the terminology functions within Pathling against arbitrary Spark dataframes.

Resolves #601, #602 and #603.


johngrimes:
@piotrszul There are a couple of remaining issues with this that I need your help with:

- The DataFrame objects that we return do not seem to be compatible with `toPandas()`. We're getting an error: `AttributeError: 'SparkSession' object has no attribute '_conf'`
- There is a problem with Databricks and the terminology functions (`PathlingStrategy`). We get this error upon trying to load the code in that environment: `java.lang.IncompatibleClassChangeError: class au.csiro.pathling.sql.MapWithPartitionPreview can not implement org.apache.spark.sql.catalyst.plans.logical.UnaryNode, because it is not an interface (org.apache.spark.sql.catalyst.plans.logical.UnaryNode is in unnamed module of loader 'app')`

johngrimes:
@piotrszul Thanks for your help with fixing those issues! 🎉

Let me know when you have finished reviewing and I will start shaping up a v5.3.0 branch.

lib/python/pathling/functions.py
```diff
@@ -0,0 +1,41 @@
+from pyspark.sql import Column
+from pyspark.sql.functions import lit, struct
+
+from pathling.coding import Coding
+
+
+def to_coding(coding_column: Column, system: str, version: str = None):
+    """
```
piotrszul, line 8: Maybe it would be also useful to create codings where system is a column not literals.
This could either be done with a function like:

```def to_coding(coding: Union[Column, str], system:Union[Column, str] ... ) ```

and then testing of the actual type. The only problem is that this is somewhat inconsistent with spark where passing a string where a column is expected is interpreted as the name of the column.

So it could also be:

```def to_coding(coding_col: Column, system_col:Optional[Column] = None,  system_lit:Optional[str]  ) ```

and then call it like: 

```to_coding(col("code"), system_lit = 'http://loinc.org") ```

File: library-api/src/main/java/au/csiro/pathling/library/PathlingContext.java [hunk 2]
```diff
@@ -315,4 +360,75 @@ public Dataset<Row> encodeBundle(@Nonnull final Dataset<Row> stringBundlesDF,
     return encodeBundle(stringBundlesDF, resourceName, FhirMimeTypes.FHIR_JSON);
   }
 
+  @Nonnull
+  public Dataset<Row> memberOf(@Nonnull final Dataset<Row> dataset,
+      @Nonnull final Column coding, @Nonnull final String valueSetUri,
+      @Nonnull final String outputColumnName) {
+
+    final Column codingArrayCol = when(coding.isNotNull(), array(coding))
```
piotrszul, line 368: I think it would be useful to refactor this functionality so that it can also be reused in MemebrOf function.

Maybe something like 

``` 
class TerminologyFunctions {

    private final TerminologyServiceFactory terminologyServiceFactory;

    @Nonnull
    public Dataset<Row> memberOf(@Nonnull final Dataset<Row> dataset,
      @Nonnull final Column coding, @Nonnull final String valueSetUri,
      @Nonnull final String outputColumnName);
      
      // other terminology functions.```

File: library-api/src/main/java/au/csiro/pathling/library/PathlingContext.java [hunk 3]
```diff
@@ -11,31 +11,70 @@
...
 
 /**
- * Main entry point to Pathling Java API
+ * A class designed to provide access to selected Pathling functionality from non-JVM language
+ * libraries.
+ *
+ * @author Piotr Szul
+ * @author John Grimes
  */
 public class PathlingContext {
```
piotrszul, line 74: I am wondering if we actually need the `PathlingContext` in this form both in Java and in Python.
Firstly we do not actually need to store SparkSession because all our functions take a dataframe as an argument an we could get the session from it.
Secondly maybe it would be better to split this into to "contexts": one specialised for encoding (only needing the relevant argument) and one specialized terminology.

maybe something like this:

``` 
import pathling.Encoder
import pathling.Terminlogy

encoder = Encoder("R4", max_nesting_level = 2)
encoder.encode_json( ....)


term = Temininology("http://server.url).
term.member_of (....)

```

piotrszul (commented):
Looks good to me!
I added a few ideas for improvements in the comments, which I am happy to discuss/help out with.

fhir-server/src/test/java/au/csiro/pathling/test/integration/modification/ModificationTest.java
```diff
@@ -78,7 +79,8 @@ static void registerProperties(@Nonnull final DynamicPropertyRegistry registry)
   void setUp() throws IOException {
     databaseDirectory = new File(
         configuration.getStorage().getWarehouseUrl().replaceFirst("file://", ""), "default");
-    final Path source = new File(PARQUET_PATH).getAbsoluteFile().toPath();
+    final Path source = Path.of(
+        String.join("/", WAREHOUSE_URL, DATABASE_NAME).replaceFirst("file://", ""));
     final Path destination = databaseDirectory.toPath();
```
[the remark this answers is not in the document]
johngrimes: Fixed in b064dbd1071ae1c5ff8990e0719cac27f264d46c

fhir-server/src/test/java/au/csiro/pathling/test/system/DockerImageTest.java
```diff
@@ -453,4 +456,15 @@ static class ClientCredentialsResponse {
...
   }
 
+  @Nonnull
+  private static String getRequiredProperty(@Nonnull final String key) {
+    final String value = System.getProperty(key);
+    final String errorMessage = "Property " + key + " must be provided";
+    if (value == null || key.isBlank()) {
+      log.error(errorMessage);
+      throw new IllegalArgumentException(errorMessage);
+    }
```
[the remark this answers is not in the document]
johngrimes: Fixed in 97a559095786812790e88759d1dac1a8b9c50949

lib/python/pathling/context.py
```diff
@@ -30,76 +36,166 @@ def spark(self) -> SparkSession:
 
     @classmethod
     def create(cls, spark: SparkSession = None,
-               fhirVersion: Optional[str] = None,
-               maxNestingLevel: Optional[int] = None,
-               enableExtensions: Optional[bool] = None,
-               enabledOpenTypes: Optional[Sequence[str]] = None) -> "PathlingContext":
+               fhir_version: Optional[str] = None,
```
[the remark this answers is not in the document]
johngrimes: Fixed in 9624ab7699ecdf657d91967ab86326e00fa46d77

library-api/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
```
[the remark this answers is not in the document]
johngrimes: We are sharing some of the test helper code between the modules.

lib/python/pathling/functions.py
johngrimes: I agree that this would be pretty easy to do, but I don't think that we have this requirement yet.

I think the scenario is that we have non-FHIR data with a system-specific column that contains a FHIR URI for the system (which wouldn't be the case in the case of say, OMOP, they use different URIs that would need to be translated anyway).

I think we should implement this feature on-demand.

lib/python/pathling/functions.py
```diff
@@ -0,0 +1,41 @@
...
+    system_column = lit(system).alias('system')
+    version_column = lit(version).alias('version')
+    display_column = lit(None).alias('display')
+    user_selected_column = lit(None).alias('userSelected')
+    return struct(id_column, system_column, version_column, coding_column.alias("code"),
+                  display_column, user_selected_column)
+
+
+def to_coding_literal(coding: Coding):
+    """
```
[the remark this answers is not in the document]
johngrimes: Resolved in c9943168a1355f5e824f228c61740319310036d1

library-api/src/main/java/au/csiro/pathling/library/PathlingContext.java
johngrimes: I agree that we shouldn't need to store the `SparkSession` in the `PathlingContext`. I have removed it from the Java class here: a676d227b5232220bb00605eb3dc3569014a8c28

Not too sure about making the same change in the Python `PathlingContext`, as that might impact your implementation of `PathlingContext._wrap_df`.

I think I might need some more convincing about the "different contexts for different use cases" thing - I think that might make things more complicated. I'm pretty happy with the fact that we can instantiate this thing and use it for either use case, all of the arguments are optional.

I can see a case for a more sophisticated solution for passing configuration to the PathlingContext as we build more functionality into it. The named arguments feature of Python means that this is not so bad even as it is, I think.

library-api/src/main/java/au/csiro/pathling/library/PathlingContext.java
johngrimes: Resolved this for memberOf, translate and subsumes in: d8e074150993d4721bd6f95c82d6db9cf946e44f, dccf4bfa3198e040cb3a168d80f413582ad014df and 8b3df9369113fd5a3cb1b314862d99e7e9df226a

PR #1108, opened 2022-12-05
v6.1.0

johngrimes:


johngrimes:
@piotrszul Just taking a look at the performance of a query in which all terminology requests are already cached.

Seems like almost half the time is spent doing JSON parsing and invoking the interceptor mechanism.

image

johngrimes:
I worked out that logging affects performance significantly.

I think we should only register the client logging interceptor if verbose request logging is on.

johngrimes:
I've got a working implementation using Infinispan instead of the default cache storage, using the [soft-index file store](https://infinispan.org/docs/stable/titles/configuring/configuring.html#file-stores_persistence). Seems to work well.

I'll form this up into a pull request.

johngrimes:
Ok, I can now see that you have already enabled disk-based caching - although it's not immediately obvious how to configure it.

johngrimes:
There seems to be some sort of problem with the disk storage of the caching mechanism as implemented.

Upon restarting, it seems to miss on the cache and generate new files.

I'm also a bit concerned that it creates a file for each unique request - I'm not sure if 100,000s - 1,000,000s of files will be a problem at some point.

johngrimes:
I think `subsumes` / `subsumedBy` need to be made more tolerant. I think we need to treat a 400 series error as a `false` for the purposes of these operations. A 400 series error can occur if the code is not found, or if the code is invalid.

The same logic probably needs to be applied to `memberOf` (`validate-code`).

In the case of `property` / `designation` / `display` (`lookup`), I think a 400 series error should simple yield no value.

johngrimes:
@piotrszul I think it might be easier to merge `issue/1066` and `caching-updates` to `release/6.0.1` sooner rather than later.

We can continue to review and refine on the feature branch, and you can merge back to your in-progress branch when it suits you.

johngrimes:
I think `spark.sql.adaptive.advisoryPartitionSizeInBytes` might be one way of increasing the parallelism in some queries - however I haven't been able to use it to get an overall performance improvement.

johngrimes:
I tried substituting the HTTP caching for a simple memoization of the call to the terminology service, and it resulted in a big speedup (~70%) for the fully cached case, but little to no speedup for the uncached case.

johngrimes:
@piotrszul I've now merged the `issue/1066` and `caching-updates` to `release/6.0.1`.

johngrimes:
@piotrszul I think this is pretty much ready to go with the exception of one problem. I am having trouble getting the library to work on Databricks.

There seems to be a conflict relating to the package `com.github.ben-manes.caffeine:caffeine`.

We have this package as a transitive dependency (through Infinispan), at version 2.9.3. Databricks (12) has version 2.3.4 of this library on the classpath.

This causes the following error:

```Caused by: org.infinispan.commons.CacheException: java.lang.NoSuchMethodError: 'java.lang.Object com.github.benmanes.caffeine.cache.Policy.getIfPresentQuietly(java.lang.Object)' at org.infinispan.interceptors.impl.InvocationContextInterceptor.rethrowException(InvocationContextInterceptor.java:135) at org.infinispan.interceptors.impl.InvocationContextInterceptor.lambda$new$0(InvocationContextInterceptor.java:61) at org.infinispan.interceptors.ExceptionSyncInvocationStage.andExceptionally(ExceptionSyncInvocationStage.java:40) at org.infinispan.interceptors.BaseAsyncInterceptor.invokeNextAndExceptionally(BaseAsyncInterceptor.java:133) at org.infinispan.interceptors.impl.InvocationContextInterceptor.visitCommand(InvocationContextInterceptor.java:89) at org.infinispan.interceptors.impl.AsyncInterceptorChainImpl.invoke(AsyncInterceptorChainImpl.java:247) at org.infinispan.cache.impl.InvocationHelper.doInvoke(InvocationHelper.java:297) at org.infinispan.cache.impl.InvocationHelper.invoke(InvocationHelper.java:101) at org.infinispan.cache.impl.CacheImpl.get(CacheImpl.java:542) at org.infinispan.cache.impl.CacheImpl.get(CacheImpl.java:536) at org.infinispan.cache.impl.AbstractDelegatingCache.get(AbstractDelegatingCache.java:439) at org.infinispan.cache.impl.EncoderCache.get(EncoderCache.java:683) at au.csiro.pathling.terminology.caching.CachingTerminologyService.getFromCache(CachingTerminologyService.java:166)```

To reproduce this problem, you can build this branch and install the Python wheel and the "all" version of the library-api JAR. I realise that this is not exactly the same method of installation as we would use post-release, where we would instruct Databricks to install library-api from Maven. I'm not sure how to emulate this any better.

The library seems to work fine when used with a regular Spark environment, such as the one created by the Dockerfile in `lib/python` using the `jupyter/all-spark-notebook` base image.

PR #1196, opened 2023-01-12
v6.1.3

johngrimes:


johngrimes:
@piotrszul I've created some pre-release packages and loaded them into Databricks.

Seems to have fixed the problem. I didn't set `spark.driver.userClassPathFirst`.

Take a look and let me know what you think. If you are happy, I will merge this and release it as v6.1.3.

PR #1292, opened 2023-02-19
Add server cache image

johngrimes:
This change adds a Dockerfile which can be used to build a front-side cache for Pathling server.

### DIFF
--- a/deployment/cache/default.vcl
+++ b/deployment/cache/default.vcl
@@ -1,0 +1,49 @@
+// Copyright 2023 Commonwealth Scientific and Industrial Research
+// Organisation (CSIRO) ABN 41 687 119 230.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


deployment/cache/default.vcl
```diff
@@ -0,0 +1,49 @@
...
+    set beresp.http.x-original-status = beresp.status;
+    set beresp.status = 200;
+  }
+  return (deliver);
+}
+
+sub vcl_deliver {
+  set resp.status = std.integer(resp.http.x-original-status, resp.status);
+  unset resp.http.x-original-status;
+}
```
piotrszul, line 49: ChatGPT says is can be done more explicitely:

sub vcl_backend_response {
  # Cache 202 responses for 1 minute
  if (beresp.status == 202) {
    set beresp.uncacheable = false;
  }
}

johngrimes:
Made redundant by #1414.

PR #1293, opened 2023-02-19
Add server Helm chart

johngrimes:
This change adds a Helm chart that installs Pathling server to a Kubernetes cluster.


piotrszul (commented):
Looks good to me.

johngrimes:
Added in v6.4.0.

PR #1505, opened 2023-06-28
Fix problems with resolution of git.properties by giving it a unique name

### LINKED ISSUE #1187: Wrong version advertised by library terminology client
It seems like the wrong `git.properties` file is being picked up from this distribution of the software, probably due to there being multiple within the JAR.

To reproduce this, use the no-argument `PathlingContext.create()` within the Python library, and take a look at the `User-Agent` header attached to terminology requests.

johngrimes:
Resolves #1187.


piotrszul (commented):
Looks good to me!

PR #1510, opened 2023-06-29
Create library runtime jar 

### LINKED ISSUE #1443: Databricks classpath conflicts
This is for release 6.2.0, when attaching Pathling library-api with maven coordinates. 
When running on Databricks we see two kind of issues:

a) terminology functions fail to work due to a conflicting version of caffeine (required by infinispan caching). in particular the version required by pathling is newer then the one provided by Databricks. The workaround it to set `spark.executor.userClassPathFirst true` on the cluster.

b) the setting above however, leads to an exception when using delta MERGE mode, as pathing seems to bring an incompatible version of `org.apache.hadoop:hadoop-client-api:`. The workaround for this problem is to exclude this artefact in Databricks library excludes section.

Ideally we would like the Pathlling library to work without either of theses workaround. The solution my may be to replace the main jar artefact with the shaded version with all conflicting dependencies relocated and with the dependency reduced pom. 
(See hadoop client runtime jar for an inspiration).

piotrszul:
Resolves #1443

### DIFF
--- a/lib/python/MANIFEST.in
+++ b/lib/python/MANIFEST.in
@@ -1,2 +1,2 @@
 global-exclude *.py[cod] __pycache__ .DS_Store
-recursive-include target/dependency *-all.jar
+recursive-include target/dependency library-runtime-*.jar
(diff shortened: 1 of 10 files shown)


johngrimes (approved):
Looks good, let's merge to the release branch then test Maven integration.

PR #1611, opened 2023-08-11
Implement Pathling R API for sparklyr 

### LINKED ISSUE #193: R integration
This change introduces a new CRAN package for [R](https://www.r-project.org/), which allows R users to make queries to Pathling and receive the results as a native [R data frame](https://cran.r-project.org/doc/manuals/r-release/R-intro.html#Data-frames) that they can use within their R environment.

See [APIs for Python and R](https://pathling.csiro.au/docs/roadmap.html#apis-for-python-and-r).

piotrszul:
Resolves #193.


johngrimes:
To do:

- [ ] Rename from `pathlyr` to `pathling`
- [ ] Set up CI to publish to a file within GitHub releases
- [ ] Solve the problem of Spark download not working, due to 3.3.2 being missing from the default mirror and 3.3.3 not being a recognised available version in SparklyR
- [ ] Solve the problem of being able to configure in a working Hadoop installation on Windows
- [ ] Update the README to contain all of the manual installation steps, for both Windows and non-Windows environments
- [ ] `pathlyr:::ptl_spark_install()` should be exported as a public function

piotrszul:
> Solve the problem of being able to configure in a working Hadoop installation on Windows:

Both need to be set as SYSTEM (not USER) env variables:

HADOOP_HOME=c:\hadoop\hadoop-3.0.0 
PATH=c:\hadoop\hadoop-3.0.0\bin;%PATH%

Also see: https://kontext.tech/article/246/install-hadoop-300-in-windows-single-node

PR #1673, opened 2023-09-03
Added Spring Boot Actuator and metrics endpoints

### LINKED ISSUE #1671: Consider adding the Spring Boot Actuator package to the server for helpful monitoring features
The Spring Boot Actuator package (https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html) allows exposing health and telemetry endpoints. This is especially neat when deploying to Kubernetes since it supports liveness- and readiness-probes: https://spring.io/blog/2020/03/25/liveness-and-readiness-probes-with-spring-boot.

To also expose metrics in Prometheus format, an additional dependency on https://mvnrepository.com/artifact/io.micrometer/micrometer-registry-prometheus is required. This can track HTTP request durations, but I believe it doesn't yet work out-of-the-box with the HAPI FHIR Facade.

I'd be happy to add these dependencies in a PR if it seems useful to you as well! - the changes are basically the same as when the HAPI FHIR JPA Starter project introduced the same: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/337 & https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/355

chgl:
Closes #1671 .

This adds the spring-boot-starter-actuator and micrometer-registry-prometheus dependencies, which expose health and metrics endpoints.
In the default configuration, health-endpoints are exposed under `:8080/actuator/health` and metrics at ``:8080/actuator/prometheus`.

### DIFF
--- a/fhir-server/pom.xml
+++ b/fhir-server/pom.xml
@@ -122,6 +122,10 @@
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-actuator</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-test</artifactId>
       <scope>test</scope>
@@ -166,6 +170,11 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+      <version>1.11.3</version>
     </dependency>
 
     <!-- Testing -->
--- a/fhir-server/src/main/resources/application.yml
+++ b/fhir-server/src/main/resources/application.yml
@@ -6,6 +6,14 @@ spring:
   banner:
     location: classpath:banner.txt
 
+# Actuator endpoints let you monitor and interact with your application.
+# <https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints>
+management:
+  endpoints:
+    web:
+      exposure:
+        include: "health,prometheus"
+
 pathling:
   # Controls the description of this server as displayed within the FHIR CapabilityStatement.
   implementationDescription: Yet Another Pathling Server
@@ -20,7 +28,7 @@ pathling:
   spark:
     # The name that Pathling will be identified as within the Spark cluster.
     appName: pathling
-    
+
   storage:
     # The base URL at which Pathling will look for data files, and where it will save data received
     # within import requests. Can be an Amazon S3 (s3://), HDFS (hdfs://) or filesystem (file://)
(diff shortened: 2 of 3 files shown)


johngrimes:
I am thinking that perhaps the default port for this interface should be different from the default port that is used for the FHIR interface.

This is because usually you would want a different level of visibility of this service, e.g. it might only need to be accessible by a monitoring component within the cluster, not exposed externally.

johngrimes:
I've made this change in 280fef19acd86130bb923cab3d8fabf845929fac, let me know what you think.

chgl:
That's definitely a great change - I've defaulted to the same behavior in the HAPI FHIR Helm chart as well: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/charts/hapi-fhir-jpaserver/templates/deployment.yaml#L98. 

One more thing (we can of course always override any of these properties at runtime), but what do you think about the `add-additional-paths` property: https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.kubernetes-probes setting. It would expose the metrics at a dedicated port (8081) but the liveness and readiness probes are kept at `:8080/livez` and `:8080/readyz`. Alternatively, to keep the interface FHIR-native, we could introduce a `$readyz/livez` operation at some later point.

johngrimes:
The probes could be on the management port as well right? I'm not aware of any requirement to have the probes available on the same port as the application.

I would imagine that something like this in the container spec would work well:

```yaml
startupProbe:
  httpGet:
    path: /actuator/health/liveness
    port: 8081
  periodSeconds: 5
  failureThreshold: 36
livenessProbe:
  httpGet:
    path: /actuator/health/liveness
    port: 8081
  periodSeconds: 5
  failureThreshold: 12
readinessProbe:
  httpGet:
    path: /actuator/health/readiness
    port: 8081
  periodSeconds: 5
  failureThreshold: 3
```

chgl:
Absolutely. the given rationale for using the same port as the main application:

> If your Actuator endpoints are deployed on a separate management context, the endpoints do not use the same web infrastructure (port, connection pools, framework components) as the main application. In this case, a probe check could be successful even if the main application does not work properly (for example, it cannot accept new connections). For this reason, is it a good idea to make the liveness and readiness health groups available on the main server port. This can be done by setting the following property: management.endpoint.health.probes.add-additional-paths=true

johngrimes:
Ok, that makes sense.

I'm not worried about the FHIR thing, as this would be outside the FHIR endpoint anyway (`/fhir`).

Would this be exposing only the liveness and readiness endpoints, or also the Prometheus endpoint? Are there any security implications to having this accessible externally?

chgl:
The Prometheus endpoint will only be exposed on port 8081 at /actuator/prometheus. It's only the health probes that will be available on :8080 as well. There's some docs on security: https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.security . Since we explicitly expose only health and metrics endpoints there shouldn't be anything sensitive in there.

johngrimes:
Thanks for this!

I think we should go ahead and change the default configuration as you suggest, to expose the health endpoints on 8080 and the other actuator information on 8081.

johngrimes (approved):
I think this looks good, and we can roll it in to the next release.

Thanks @chgl!

johngrimes:
@chgl Just letting you know that this change is now in v6.4.0 - check it out and let us know what you think.

chgl:
@johngrimes , awesome, thanks! Works great! I've already updated my helm chart to use the new endpoints for liveness/readiness and metrics.

PR #1781, opened 2024-02-19
Implement FHIR Bulk Export client.

### LINKED ISSUE #1776: Bulk FHIR client
It would be great if there was a bulk FHIR client implementation within Pathling.

This client would be accessible through the library API, so that you could configure and run it to retrieve a set of NDJSON files from a server and save them to a Hadoop-compatible file location.

The operation should gracefully handle all foreseeable error states from the server, and retry requests upon any transient errors.

piotrszul:
Resolves #1776


bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
```diff
@@ -0,0 +1,278 @@
...
+  @Nonnull
+  @Builder.Default
+  Level level = new SystemLevel();
+
+  /**
+   * The format of the output. The value of the '_outputFormat' query parameter.
+   */
+  @Nullable
+  @Builder.Default
+  String _outputFormat = null;
```
johngrimes, line 149: What happens if I request something other than NDJSON, and receive it?

Do we have assumptions elsewhere that will be broken by this?

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
```diff
@@ -0,0 +1,278 @@
...
+  @Builder.Default
+  List<String> _typeFilter = Collections.emptyList();
+
+
+  /**
+   * The patient(s) to include in the export. The value of the 'patient'  parameter.
+   */
+  @Nonnull
+  @Builder.Default
+  List<Reference> patient = Collections.emptyList();
```
johngrimes, line 187: Is this the full list of parameters that I can use? Is it possible to pass parameters that are not in this class, e.g. `includeAssociatedData`?

bulkclient/src/main/java/au/csiro/pathling/export/BulkExportClient.java
```diff
@@ -0,0 +1,389 @@
...
+        final ExecutorServiceResource executorServiceResource = createExecutorServiceResource()
+    ) {
+      final BulkExportTemplate bulkExportTemplate = new BulkExportTemplate(
+          new BulkExportAsyncService(httpClient, URI.create(fhirEndpointUrl)),
+          asyncConfig);
+      final UrlDownloadTemplate downloadTemplate = new UrlDownloadTemplate(httpClient,
+          executorServiceResource.getExecutorService());
+
+      final BulkExportResult result = doExport(fileStore, bulkExportTemplate, downloadTemplate);
+      log.info("Export successful: {}", result);
```
johngrimes, line 260: Should we issue a [Bulk Data Delete Request](https://build.fhir.org/ig/HL7/bulk-data/export.html#bulk-data-delete-request) once the download has successfully completed (if the server supports it)?

bulkclient/src/test/java/au/csiro/pathling/example/RunBulkExport.java
```diff
@@ -0,0 +1,314 @@
...
+    final String fhirEndpointUrl = "https://bulk-data.smarthealthit.org/eyJlcnIiOiIiLCJwYWdlIjoxMDAwMCwiZHVyIjoxMCwidGx0IjoxNSwibSI6MSwic3R1Ijo0LCJkZWwiOjAsInNlY3VyZSI6MX0/fhir";
+    final Instant from = Instant.parse("2020-01-01T00:00:00.000Z");
+    // Bulk Export Demo Server
+    final String outputDir = "target/export-" + Instant.now().toEpochMilli();
+
+    System.out.println(
+        "Exporting" + "\n from: " + fhirEndpointUrl + "\n to: " + outputDir + "\n since: " + from);
+
+    final String clientId = IOUtils.toString(
+        RunBulkExport.class.getClassLoader().getResourceAsStream("auth/bulk_rs384_clientId.txt"));
```
johngrimes, line 203: This seems to be a very common call, perhaps we should pull this out into a helper method.

johngrimes (changes requested):
Thanks for all of your work on this @piotrszul!

I've gone through it and come up with some feedback for you to take a look at.

bulkclient/src/main/java/au/csiro/pathling/auth/JWTUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    } catch (final JOSEException | InvocationTargetException | IllegalAccessException e) {
+      throw new RuntimeException("Exception while creating Algorithm for: " + jwtAlgoName, e);
+    }
+  }
+
+  private static String toAuto0AlgoName(@Nonnull final String jwkAlgorithm) {
+    return jwkAlgorithm
+        .replace("RS", "RSA")
+        .replace("ES", "ECDSA")
+        .replace("HC", "HMAC");
```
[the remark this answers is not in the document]
piotrszul: It should actually be "HS" not "HC" as if HS256.
it's one of the symmetric signing algorithms supported by auth0. 
Added for completness as it's supported by 'Algorithm' class but we do not use that since we only expose asymmetric algorithms.

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportTemplate.java
```diff
@@ -0,0 +1,294 @@
...
+
+    @Nonnull
+    private PoolingState handleContinueResponse(
+        @Nonnull final AcceptedAsyncResponse acceptedResponse) {
+      log.debug("Pooling: progress: '{}', retry-after: {}",
+          acceptedResponse.getProgress().orElse("na"),
+          acceptedResponse.getRetryAfter().map(RetryValue::toString).orElse("na"));
+      final Duration timeToSleep = computeTimeToSleep(acceptedResponse.getRetryAfter(),
+          config.getMinPoolingDelay());
+      log.debug("Pooling: Sleeping for {} ms", timeToSleep.toMillis());
```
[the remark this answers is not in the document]
piotrszul: Yes.
AsyncConfig:
  /**
   * The maxium delay between two consecutive status requests. This may override the value returned
   * from the server in the 'retry-after' header.
   */
  @Builder.Default
  Duration maxPoolingDelay = Duration.ofSeconds(60);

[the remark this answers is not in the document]
johngrimes: Great, thanks.

bulkclient/src/test/java/au/csiro/pathling/example/RunBulkExport.java
piotrszul: OK

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
piotrszul: Not at the moment. The exported files are 'blackboxes' we just save what the server provides without any interpretation. 
The only potentially correlated thing is the:
BulkExportClient:
  /**
   * The extension to use for the output files.
   */
  @Nonnull
  @Builder.Default
  String outputExtension = "ndjson";
  
  
  This will need to be manually adjusted to the actual format.
  
  This may become a bit more complicated if we add support for downloading 'error' URLs from the response as these should be JSON operation outcomes.

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
piotrszul: Yes, this is the full list.
Actually whatever is defined in `BulkExportClient` but this is then mapped to the `BulkExportRequest`.
I think `includeAssociatedData` is the only one missing from the spec.

bulkclient/src/main/java/au/csiro/pathling/export/BulkExportClient.java
piotrszul: The spec say that we MAY issue it to either cancel the request in progress or to cleanup.
So I guess it depends what we want to consider to be required for the first release of the bulk client.
I this this might be something to be added  subsequent releases.

bulkclient/examples/bulk_export_smilecdr.py
```diff
@@ -0,0 +1,67 @@
...
+    timeout = timedelta(minutes=30),
+    auth_config=auth_config(
+        auth_enabled=True,
+        auth_token_endpoint="https://aehrc-cdr.cc/smartsec_r4/oauth/token",
+        auth_client_id="pathling-bulk-client",
+        auth_client_secret=clientSecret,
+        auth_scope="system/*.read",
+    ),
+).export()
+print(f"TransactionTime: {as_fhir_instant(result.transaction_time)}")
```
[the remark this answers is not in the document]
piotrszul: Yes, this should be resolved by JSR-380 validation of config. 
I was having some issues with the making it work (in this project - some dependency issues).

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
johngrimes: I think we should add this to make the list complete.

bulkclient/src/main/java/au/csiro/pathling/export/BulkExportClient.java
johngrimes: I think we should add this to mitigate the problem of exported data accumulating in the source server, which has been a problem that we have encountered with HAPI JPA.

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
johngrimes: I think we can leave this as is for now, as NDJSON is the only current format.

bulkclient/src/main/java/au/csiro/pathling/export/BulkExportClient.java
```diff
@@ -0,0 +1,389 @@
...
+    }
+    final BulkExportResponse response = bulkExportTemplate.export(buildBulkExportRequest(),
+        TimeoutUtils.toTimeoutAfter(timeoutAt));
+    log.debug("Export request completed: {}", response);
+
+    final List<UrlDownloadEntry> downloadList = getUrlDownloadEntries(response, destinationDir);
+    log.debug("Downloading entries: {}", downloadList);
+    final List<Long> fileSizes = downloadTemplate.download(downloadList,
+        TimeoutUtils.toTimeoutAfter(timeoutAt));
+    final FileHandle successMarker = destinationDir.child("_SUCCESS");
```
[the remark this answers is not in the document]
johngrimes: This is a feature of HDFS and only happens when using this through the Pathling library.

bulkclient/src/main/java/au/csiro/pathling/auth/ClientAuthMethod.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.Value;
+import org.apache.http.client.HttpClient;
+
+/**
+ * Authentication method for one of the FHIR SMART client authentication profiles.
+ */
+
+public interface ClientAuthMethod {
```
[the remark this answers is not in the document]
piotrszul: Refactored.

bulkclient/src/main/java/au/csiro/pathling/auth/AuthConst.java
```diff
@@ -0,0 +1,68 @@
...
+
+package au.csiro.pathling.auth;
+
+import lombok.experimental.UtilityClass;
+
+/**
+ * Constants for authentication.
+ */
+@UtilityClass
+public class AuthConst {
```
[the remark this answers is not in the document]
piotrszul: Reafactored.

bulkclient/src/main/java/au/csiro/pathling/export/ws/BulkExportRequest.java
piotrszul: Implemented.

bulkclient/src/main/java/au/csiro/pathling/export/BulkExportClient.java
piotrszul: Implemented.

bulkclient/examples/bulk_export_smilecdr.py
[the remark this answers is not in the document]
piotrszul: Added JRS-380 validation for the client configuration and it's  subconfig beans.

johngrimes:
We ended up shipping this client within a separate repo: https://github.com/aehrc/fhir-bulk-java

PR #1775, opened 2024-01-16
Array representations within FHIRPath queries

piotrszul:
Resolves #1759

PR #1786, opened 2024-04-10
Implementing full encoding and decoding support for References

### LINKED ISSUE #1784: Reference.identifier is not encoded
**Describe the bug**
We create resources that set both `Reference.reference` and `Reference.identifier` for references to `Patient` and `Encounter` resources.

For example:

```json
{
  "resourceType": "Condition",
  "id": "123",
  "subject": {
    "reference": "Patient/456",
    "identifier": {
      "type": {
        "coding": [
          {
            "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
            "code": "MR"
          }
        ]
      },
      "system": "https://fhir.example.com/identifiers/mrn",
      "value": "123456789"
    }
  }
}
```

However, while `Condition.subject.reference` is encoded correctly, the `Condition.subject.identifier` is ignored. The following SQL:

```sql
SELECT subject
FROM delta.`s3a://pathling-warehouse/default/Condition.parquet`
```

returns only:

```json
{"reference":"Patient/456","display":null,"_fid":447035444}
```

The FHIR spec https://hl7.org/fhir/R4B/references.html#logical states:

> When both logical and literal references are provided, the literal reference is preferred. Applications processing the resource are allowed - but not required - to check that the identifier matches the literal reference, if they understand how to resolve the logical reference.

So this does make sense, but it would still be nice to have it preserved.

**To Reproduce**

1. Create a resource referencing a different resource with both (or just setting the `Reference.identifer` might do the trick), like in the example above
2. POST it to the Pathling Server to have it encoded
3. Observe that the encoded resource no longer contains the `Reference.identifier`

**Expected behavior**

Both the `Reference.reference`  and `Reference.identifier` should be preserved when encoding.

This is definitely nothing critical, it's just a helpful extra-information we include in resources to make identification easier.

piotrszul:
Resolves #1784

### DIFF
--- a/encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
+++ b/encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
@@ -24,10 +24,8 @@
 package au.csiro.pathling.encoders
 
 import au.csiro.pathling.encoders.ExtensionSupport.{EXTENSIONS_FIELD_NAME, FID_FIELD_NAME}
-import au.csiro.pathling.encoders.QuantitySupport.{CODE_CANONICALIZED_FIELD_NAME, VALUE_CANONICALIZED_FIELD_NAME}
 import au.csiro.pathling.encoders.SerializerBuilderProcessor.{dataTypeToUtf8Expr, getChildExpression, objectTypeFor}
-import au.csiro.pathling.encoders.datatypes.{DataTypeMappings, DecimalCustomCoder}
-import au.csiro.pathling.encoders.terminology.ucum.Ucum
+import au.csiro.pathling.encoders.datatypes.DataTypeMappings
 import au.csiro.pathling.schema.SchemaVisitor.isCollection
 import au.csiro.pathling.schema._
 import ca.uhn.fhir.context.BaseRuntimeElementDefinition.ChildTypeEnum
@@ -36,7 +34,7 @@ import org.apache.spark.sql.catalyst.expressions.objects.{ExternalMapToCatalyst,
 import org.apache.spark.sql.catalyst.expressions.{BoundReference, CreateNamedStruct, Expression, If, IsNull, Literal}
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
-import org.hl7.fhir.instance.model.api.{IBaseDatatype, IBaseHasExtensions, IBaseResource}
+import org.hl7.fhir.instance.model.api.{IBaseDatatype, IBaseHasExtensions, IBaseReference, IBaseResource}
 import org.hl7.fhir.r4.model.{Base, Extension, Quantity}
 import org.hl7.fhir.utilities.xhtml.XhtmlNode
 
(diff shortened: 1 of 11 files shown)


johngrimes (changes requested):
I don't seem to be able to traverse into `identifier` within a `Reference` within the context of a FHIRPath query, even though this field has been encoded in the schema.

```python
data.extract("Condition", ["id", "subject.identifier.value"])
```

Would you mind giving this a go and seeing if you can reproduce?

johngrimes:
This is the error that occurs if you try to traverse to `assigner` within an `Identifier` within a `Reference`:

```
pyspark.errors.exceptions.captured.AnalysisException: [FIELD_NOT_FOUND] No such struct field `assigner` in `id`, `use`, `type`, `system`, `value`, `period`, `_fid`.
```

I think this is fine for now.

encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
```diff
@@ -226,16 +224,22 @@ private[encoders] object SerializerBuilderProcessor {
...
-      } else {
+    field match {
+      case p: RuntimeChildPrimitiveDatatypeDefinition if p.getMax == 1 && p
+        .getElementName != "div" =>
+        if ("reference" == p.getElementName && classOf[IBaseReference]
+          .isAssignableFrom(p.getField.getDeclaringClass)) {
+          // special case for subclasses of IBaseReference
+          // for some obscure reason the accessor getReferenceElement returns IdType rather than StringType and 
+          // getReferenceElement_ needs to be used instead.
+          "getReferenceElement_"
```
johngrimes, line 235: If we are using `getReferenceElement_`, should we be checking if it is assignable from `Reference`, not `IBaseReference`?

fhir-server/src/test/java/au/csiro/pathling/fhirpath/parser/ParserTest.java
```diff
@@ -836,15 +836,7 @@ void testQuantityAdditionWithOverflow() {
...
-
-  @Test
-  void testTraversalToUnsupportedReferenceChild() {
-    final String expression = "reverseResolve(MedicationRequest.subject).requester.identifier";
-    final InvalidUserInputError error = assertThrows(InvalidUserInputError.class,
-        expression);
-    assertEquals("No such child: " + expression, error.getMessage());
-  }
-
+
```
johngrimes, line 839: I think we should replace this test with one that verifies that we can traverse to one of the newly encoded fields within `Reference`.

fhirpath/src/test/java/au/csiro/pathling/fhirpath/parser/FhirpathTest.java
```diff
@@ -0,0 +1,115 @@
...
+  void testTraversalIntoReferenceIdentifier() {
+    withResources(
+        new Condition()
+            .setSubject(new Reference().setIdentifier(new Identifier().setValue("value")))
+            .setId("001")
+    );
+    assertThatResultOf(ResourceType.CONDITION, "subject.identifier.value")
+        .selectResult()
+        .hasRows(RowFactory.create("001", "value"));
+  }
```
johngrimes, line 113: Ok, I can see that you have actioned the previous comment here. 

I like the idea, but should we also bring the other tests over to this new scheme?

johngrimes (changes requested):
Thanks @piotrszul! The traversal problem is now fixed for me.

I've added a few further items of feedback.

encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
piotrszul: Possibly. 
`Reference` is a FHIR version specific class, so if we were to use `r4.Reference` we technically (to keep the basic SerializerBuilder version independent) should delegate this to R4TypeMappings.
Using 'IBaseReference' is a compromise as all the Reference types inherit from IBaseReference and starting from 'stu3'  have the `getReferenceElement_()` method.

fhirpath/src/test/java/au/csiro/pathling/fhirpath/parser/FhirpathTest.java
piotrszul: Possibly, but I think that should be a separate task and maybe somehow related to the testing of columnar representation. We my though keep using the new test for testing new features/changes in the parser.
Also, we should be able to test that in the existing ParserTest even though we do not have any instances of  non- empty Reference.identifiers  - but that should not matter in this case.

fhir-server/src/test/java/au/csiro/pathling/fhirpath/parser/ParserTest.java
piotrszul: Yes, added in the new test format.

encoders/src/main/scala/au/csiro/pathling/encoders/SerializerBuilder.scala
johngrimes: Ok, thanks.

I think this solution is fine as long as we have test coverage that verifies that it works.

fhirpath/src/test/java/au/csiro/pathling/fhirpath/parser/FhirpathTest.java
johngrimes: Ok. I think this might come under general refactoring of the testing strategy within #1775.

johngrimes (approved):

PR #1871, opened 2024-04-30
Moving benchmark execution to a forked VM and adding spring context cleanup

piotrszul:
Moving benchmark execution to a forked VM (from the maven one) and adding the explicit cleanup of the application context(s) created for the spring autowired benchmarks.

This should provide better isolation of the benchmark run from the maven build process.


johngrimes (approved):
Thanks @piotrszul!

PR #1876, opened 2024-05-02
Changing the SPARK encoding of empty non-primitive elements in HAPI objects

piotrszul:
Changing the SPARK encoding of empty non-primitive elements in HAPI objects to produce NULL values rather than empty structs/arrays.

For example the encoding of:
```
"identifier" : [
  {
    "use" : "official",
    "system" : "https://github.com/synthetichealth/synthea",
    "value" : "d2120f2d-bc32-096d-d854-1a1e73a05f0d"
  }
],
```

produce the row:

```
Row(identifier=[Row(id=None, use='official', type=None, system='
[https://github.com/synthetichealth/synthea'](https://github.com/synthetichealth/synthea%27)
, value='e4fc9c8b-66b7-7f4b-a033-39acfc7633fa', period=None, assigner=None)])
```
where composite elements such aas period or assigner as set to NULL (None).

The same also applies to empty elements with `many` cardinality. They produce NULL rather then an empty `array()`.

### DIFF
--- a/encoders/src/main/scala/au/csiro/pathling/encoders/Expressions.scala
+++ b/encoders/src/main/scala/au/csiro/pathling/encoders/Expressions.scala
@@ -113,6 +113,54 @@ case class InstanceOf(value: Expression,
 
 }
 
+
+/**
+ * Gets value of an element of a HAPI object. This is could be composed from `If` and `Invoke`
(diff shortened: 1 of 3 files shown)


johngrimes (approved):
Looks good to me!

PR #2062, opened 2025-01-15
Add Parquet and Delta import

### LINKED ISSUE #2061: Enable import of Parquet and Delta files
Currently the server import operation allows for the import of NDJSON only.

It would be great if this could also allow for the import of Pathling-produced Parquet and Delta tables.

johngrimes:
Resolves #2061.


fhir-server/src/main/java/au/csiro/pathling/update/ImportExecutor.java
```diff
@@ -137,6 +140,13 @@ public OperationOutcome execute(@Nonnull @ResourceParam final Parameters inParam
           .map(param -> ImportMode.fromCode(
               ((CodeType) param.getValue()).asStringValue()))
           .orElse(ImportMode.OVERWRITE);
+
+      // Get the serialized resource type from the source parameter.
+      final String format = sourceParam.getPart().stream()
```
jkiddo, line 145: Sugar on top, but this could be put into an enum

jkiddo (approved):

fhir-server/src/main/java/au/csiro/pathling/update/ImportExecutor.java
johngrimes: Fixed in 6a0b06b

Issue #2307, opened 2025-05-12
Release 8.0 descope: remove modules


piotrszul:
Related to: #2276

Remove the code and module out of scope for version 8.0

PR #2314, opened 2025-05-14
Release 8.0 descope: remove server and other modules.

piotrszul:
Resolves:  #2307


johngrimes (approved):
This looks good to me, I think we should merge this and proceed with the extract, aggregate and FHIRPath functionality aspects of this.

PR #2375, opened 2025-06-05
Add bulk client feature to release 8 branch

### LINKED ISSUE #2359: Add bulk client feature to release 8 branch
The bulk client is a new feature that was merged to the main branch since we have been working on v8.

See #2162.

johngrimes:
Resolves #2359.


johngrimes:
@piotrszul Let me know if you are happy for me to merge this into v8.

piotrszul:
@johngrimes it looks good to me, but there does not seem to any tests (java, python or R) to cover this new functionality.

johngrimes:
I actually did have a go at writing some tests for this, but they were so horrible and brittle that I ended up getting rid of them. 🤷🏻

This is really just a thin layer over the top of your Java implementation, Python and R methods that collect up the arguments and pass them through the Java FFI.

johngrimes:
I ended up "testing" it with example scripts (for both R and Python) that were run against the SMART Health IT bulk data sandbox.

I guess we could turn these into tests, but it introduces an external dependency to our build if we do.

piotrszul:
In https://github.com/aehrc/pathling/tree/issue/1776 (this was the initial branch for BulkExportClients).
There are BulkClients test with Wiremock for Java and another mock server for python.
(BulkExportClientWiremockTest.java,  mockserver.py, test_bulkexport.py).
Perhaps these can be used to mock the depedency on the extrenal bulk export server.

piotrszul:
Or perhaps we could just test the python layer checking that the java BulkClientExport was correctly setup with the python API (the BulkClientExport are set in the way consistent with the python api arguments).

johngrimes:
I've added a task to look at the tests in #2376.

PR #2377, opened 2025-06-06
Update reference implementation test config to match v8 scope

### LINKED ISSUE #2374: Update reference implementation test config to match v8 scope
We have a file (`fhirpath/src/test/resources/fhirpath-js/config.yaml`) that we use to mark tests from the fhirpath.js test suite that we don't pass yet.

Since we cut down the scope of the implementation, we haven't updated this file to reflect the new scope and ticket up the new gap between the v8 implementation and full coverage of FHIRPath.

Once we have done this we will be able to re-enable the fhirpath.js test suite and improve the coverage of the functionality within v8.

johngrimes:
Resolves #2374.


johngrimes:
@piotrszul Why do we have to skip this test?

`3.2_paths.yaml` - `Observation value with an R4 FHIR model`

How is this different to other tests that have paths that are not in the resource definition?

johngrimes:
@piotrszul I'm also trying to remember the rationale for this exclusion:

```yaml
  - title: Subsetting exclusions
    glob: "fhirpath-js/cases/5.3_subsetting.yaml"
    exclude:
      - title: "Primitive attributes"
        type: ???
        any:
          - "** first for collection with null values"
```

johngrimes:
Ok, I think I remember this one now - support for the primitive attribute/extension structure is required for this.

PR #2455, opened 2025-07-15
Add WireMock tests to bulk client functionality

### LINKED ISSUE #2376: Add WireMock tests to bulk client functionality

piotrszul:
Resolves #2376.

Also make improvements to the bulk() datasource API, including support for temporary directory as the output_dir.

### DIFF
--- a/lib/python/pathling/spark.py
+++ b/lib/python/pathling/spark.py
@@ -1,0 +1,89 @@
+#  Copyright 2025 Commonwealth Scientific and Industrial Research
+#  Organisation (CSIRO) ABN 41 687 119 230.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


lib/python/pathling/spark.py
```diff
@@ -0,0 +1,89 @@
...
+        The directory itself is not created, only the path is returned.
+        
+        :param infix: String to insert between the base directory and the UUID (default: "app-tmp").
+        :param def_temp_base: Fallback base directory if `hadoop.tmp.dir` is not set in Hadoop configuration (default: "/tmp").
+        :param qualified: If True, returns a fully qualified Hadoop path; if False, returns a raw path string.
+        :return: String representing the unique temporary directory path.
+        """
+        base_tmp = self._hadoop_conf.get("hadoop.tmp.dir") or def_temp_base
+        uuid_suffix = str(uuid.uuid4())
+        tmp_path = f"{base_tmp}/{infix}-{uuid_suffix}"
```
johngrimes, line 56: Will this work in Windows?

johngrimes (commented):
Thanks @piotrszul, I just had one query about the HDFS stuff.

lib/python/pathling/spark.py
piotrszul: It should/might is spark/hadoop is configured corectly becasue it uses hadoop's Path() class which even on windows uses forward slash ('/') as separator and the 'hadoop.tmp.dir' should be set to something valid on windows.

johngrimes: Yes, pathling-windows.it.csiro.au

PR #2464, opened 2025-07-22
Cleaning-up obsolete and unused code.

piotrszul:
Resolves #3213.


piotrszul:
@johngrimes 
I have purged I believe most of the obvious obsolete/commented code. 
The next level could be to go through the inteliJ unused declaration inspection and remove more unused methdos/classes. Do you think it's worth doing?

johngrimes:
> @johngrimes I have purged I believe most of the obvious obsolete/commented code. The next level could be to go through the inteliJ unused declaration inspection and remove more unused methdos/classes. Do you think it's worth doing?

Yes, I think so.

PR #2482, opened 2025-09-10
Pathling 8 server

fhnaumann:
Renamed the branch to reflect the planned changes more accurately.

### DIFF
--- a/fhirpath/src/test/java/au/csiro/pathling/UnitTestDependencies.java
+++ b/fhirpath/src/test/java/au/csiro/pathling/UnitTestDependencies.java
@@ -20,6 +20,7 @@ package au.csiro.pathling;
 import au.csiro.pathling.config.QueryConfiguration;
 import au.csiro.pathling.encoders.FhirEncoders;
 import au.csiro.pathling.encoders.terminology.ucum.Ucum;
+import au.csiro.pathling.io.source.DataSource;
 import au.csiro.pathling.sql.FhirpathUDFRegistrar;
 import au.csiro.pathling.sql.udf.TerminologyUdfRegistrar;
 import au.csiro.pathling.terminology.TerminologyService;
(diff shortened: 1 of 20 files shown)


fhnaumann:
Demonstration:
An AidBox server is deployed to Kubernetes. A small script pulls data from it (using bulk export?) and converts the ndjson files to parquet files (delta tables). The developed pathling-server uses that data as input and it can be requested to perform bulk export on it. Requests are made through a web client. Every component is dockerized, packaged using helm charts and then deployed to Kubernetes.

In the future, additional technologies such as databricks may be used. Also authentication and authorization may be performed through the web client to the pathling-server.

use the auth in the createTag method, but unsure what the effects are. Is there something "in" the auth object that stays the same across requests so caching still works? Anyhow, some auth information should maybe be part of the tag (if parts stay the same across requests)

johngrimes:
@fhnaumann Could you please merge main into this branch?

johngrimes:
Why are there some deleted files from the test data directory in the library API? There are other tests that rely upon this data.

johngrimes (changes requested):
Doesn't actually compile for me yet, and not passing the tests on CI.

I've added some preliminary comments anyway, I can take another look once we have a green build.

As a general comment please also take a look at the `CONTRIBUTING.md` file and make sure that everything is ticked off there.

fhnaumann:
Re-implementing the $import operation requires some changes to the API:
- `DataSourceBuilder`:
  - Add the ability to load a list of filepaths instead of just an entire directory. Proposed change: Overload the current methods which take in a directory path with new methods which take in `Collection<String> filepaths`

Change the $import parameters:
- rename `source` to `input`
- move `mode` and `format` to top-level

Align the $import to the [bulk-import manifest](https://github.com/smart-on-fhir/bulk-import/blob/master/import-manifest.md):
- only support mandatory fields (except for `inputSource`)
-

Issue #2574, opened 2026-03-18
Constants not working in python dataset view


matthewoneill7-nhs:
**Describe the bug**
When using constants in a dataset view, the query fails. 

This is because the code in view uses `constants` instead of `constant` which is in the [specification](https://sql-on-fhir.org/ig/latest/StructureDefinition-ViewDefinition.html#using-constants) and on the java side

changing the `view` method in pathling/datasource.py to take `constant` fixes the issues:

**To Reproduce**

```python
pc = PathlingContext.create(spark)
data_source = pc.read.datasets({"AuditEvent": encoded})
# where encoded is a dataframe of AuditEvents

results = data_source.view(resource="AuditEvent", 
      constants=[{"name": "requestor", "valueBoolean": "true"}]
      select = [
        {"column": [{"path": "id", "name": "event_id"}]},
        {
            "forEach": "agent.where(requestor = %requestor)",
            "column": [
                {"path": "who.display", "name": "agent_name"},
            ],
        },
    ],
)
```
produces the following traceback (truncated for relevance):
  File "site-packages/pathling/datasource.py", line 102, in view
    return self._wrap_df(jquery.execute())
                         ~~~~~~~~~~~~~~^^
  File "site-packages/py4j/java_gateway.py", line 1362, in __call__
    return_value = get_return_value(
        answer, self.gateway_client, self.target_id, self.name)
  File "site-packages/pyspark/errors/exceptions/captured.py", line 185, in deco
    raise converted from None
pyspark.errors.exceptions.captured.IllegalArgumentException: __**Unknown variable: requestor**__

**Expected behavior**
The value of the constant should be injected into the query

**Fix**
Change the argument name in DataSource.view from constants to constant and change the key in the query args to constant

```python

def view(
    self,
    resource: Optional[str] = None,
    select: Optional[Sequence[Dict]] = None,
    constant: Optional[Sequence[Dict]] = None,
    where: Optional[Sequence[Dict]] = None,
    json: Optional[str] = None,
) -> DataFrame:
    """
    Executes a SQL on FHIR view definition and returns the result as a Spark DataFrame.

    :param resource: The FHIR resource that the view is based upon, e.g. 'Patient' or
            'Observation'.
    :param select: A list of columns and nested selects to include in the view.
    :param constant: A list of constants that can be used in FHIRPath expressions.
    :param where: A list of FHIRPath expressions that can be used to filter the view.
    :param json: A JSON string representing the view definition, as an alternative to providing
[...]

PR #2575, opened 2026-03-18
bugfix: update view code params to take constant not constants

matthewoneill7-nhs:
constants gets passed to the JVM which is then ignored. This then causes any query that uses constants to fail with an 'Unknown variable' error

### DIFF
--- a/lib/python/tests/test_view.py
+++ b/lib/python/tests/test_view.py
@@ -46,26 +46,30 @@ def test_view_on_ndjson(ndjson_test_data_dir, pathling_ctx):
         ResultRow("beff242e-580b-47c0-9844-c1a68c36c5bf", "Towne435"),
     ]
 
+
 def test_view_with_constants(ndjson_test_data_dir, pathling_ctx):
     """Reproduces #2574: view with constants fails to resolve constant references in where clause."""
 
     data_source = pathling_ctx.read.ndjson(ndjson_test_data_dir)
     result = data_source.view(
         resource="Patient",
-        constant = [{"name": "filter_id", "valueUuid": "8ee183e2-b3c0-4151-be94-b945d6aa8c6d"}],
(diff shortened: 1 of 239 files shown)


matthewoneill7-nhs:
linked to #2574

lib/python/tests/test_view.py
```diff
@@ -46,27 +46,31 @@ def test_view_on_ndjson(ndjson_test_data_dir, pathling_ctx):
...
+                    {"path": "name.first().family", "name": "family_name"},
                 ]
             }
         ],
-        where=[{"path": "id = $filter_id"}],
+        where=[{"path": "id = %filter_id"}],
     )
     assert result.collect() == [
-        Row("8ee183e2-b3c0-4151-be94-b945d6aa8c6d", "Krajcik437"),
+        ResultRow("8ee183e2-b3c0-4151-be94-b945d6aa8c6d", "Krajcik437"),
```
matthewoneill7-nhs, line 70: oh no, how embarrassing. 

Apologies, I installed uv but couldn't get the tests to run

johngrimes: No problem at all!

If you let us know what you had trouble with, we will try to improve the documentation in CONTRIBUTING.md.

matthewoneill7-nhs: I made sure to have java 21 installed, and python with UV

then i ran `mvn clean install -pl lib/python -am` but it kept ending in a killed process. when i `cd`d into the python library folder and ran pytest through uv it failed with missing libraries. 

i'm not familiar with the java tooling so i don't know which knobs and dials to turn to make it less likely to crash on me. It could also be my machine, i've had wsl crash on me a fair bit while using pyspark and pathling.

johngrimes:
Released in 9.6.0.

PR #2579, opened 2026-03-26
feat: sql run operation

jkiddo:
A yolo implementation of https://github.com/aehrc/pathling/issues/2561 :

This pull request introduces support for the `$sqlquery-run` FHIR operation, enabling execution of SQL queries defined in FHIR Library resources. It adds new providers, configuration options, and parsing logic to handle SQLQuery Libraries, their parameters, and dependencies. The changes also ensure that the new operation is configurable and properly registered within the server.

**New SQL Query Operation Support:**

* Added new providers: `SqlQueryRunProvider` and `SqlQueryInstanceRunProvider` to handle both type-level and instance-level `$sqlquery-run` operations for FHIR Library resources, including parameter binding and output formatting. (`[[1]](diffhunk://#diff-7441f4e261837a0c47fe3ba576c977a207bbdc04c5aaadb30c844232af85cdebR177-R184)`, `[[2]](diffhunk://#diff-7441f4e261837a0c47fe3ba576c977a207bbdc04c5aaadb30c844232af85cdebL233-R248)`, `[[3]](diffhunk://#diff-7441f4e261837a0c47fe3ba576c977a207bbdc04c5aaadb30c844232af85cdebR277-R278)`, `[[4]](diffhunk://#diff-5acafa29251cd0573c336431c5e7f133d74246a4f80c9281d3216e3dbc3b83bbR1-R160)`)

* Registered the new SQL query run providers within the `FhirServer` initialization logic, gated by a new configuration flag. (`[server/src/main/java/au/csiro/pathling/FhirServer.javaR397-R402](diffhunk://#diff-7441f4e261837a0c47fe3ba576c977a207bbdc04c5aaadb30c844232af85cdebR397-R402)`)

**Configuration and Capability Statement:**

* Introduced a new configuration property `sqlQueryRunEnabled` to enable or disable the `$sqlquery-run` operation. (`[server/src/main/java/au/csiro/pathling/config/OperationConfiguration.javaR73-R75](diffhunk://#diff-d253d305020a8cff17b5936c099e7e2e7d731120bcc273135f5973795605eaccR73-R75)`)

* Updated the conformance/capability statement logic to advertise the `$sqlquery-run` operation when enabled. (`[server/src/main/java/au/csiro/pathling/fhir/ConformanceProvider.javaR507-R509](diffhunk://#diff-dd39c622cbec81a9a667a477be7664401ec1272b7a9d1445a2bd2a5791c8c1e5R507-R509)`)

**SQLQuery Library Parsing:**

* Added `SqlQueryLibraryParser`, which parses FHIR Library resources conforming to the SQLQuery profile, extracting SQL text, ViewDefinition dependencies, and parameter declarations. (`[server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryLibraryParser.javaR1-R143](diffhunk://#diff-1f739feb2f4b660e33bb7305adbdbbd1cbb1bbc1e9531ef64010f931193a8495R1-R143)`)

* Introduced supporting classes: `ParsedSqlQuery` (holds parsed SQL, dependencies, parameters) and `SqlParameterDeclaration` (represents a named, typed parameter). (`[[1]](diffhunk://#diff-2e4ab2cf0cdde8df9f81de693a4b02979a8b9aad69dc64988fa3690fd1523a63R1-R39)`, `[[2]](diffhunk://#diff-dcc4e4e7afa0be80e5b6930026fd9a0d0f141902bbe9983bb1b8ca72d636987aR1-R35)`)

### DIFF
--- a/server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryExecutionHelper.java
+++ b/server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryExecutionHelper.java
@@ -1,0 +1,415 @@
+/*
+ * Copyright © 2018-2026 Commonwealth Scientific and Industrial Research
+ * Organisation (CSIRO) ABN 41 687 119 230.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package au.csiro.pathling.operations.sqlquery;
+
+import au.csiro.pathling.config.ServerConfiguration;
+import au.csiro.pathling.library.io.source.QueryableDataSource;
+import au.csiro.pathling.operations.view.ResultStreamingHelper;
+import au.csiro.pathling.read.ReadExecutor;
+import au.csiro.pathling.security.PathlingAuthority;
+import au.csiro.pathling.security.ResourceAccess.AccessType;
+import au.csiro.pathling.security.SecurityAspect;
+import au.csiro.pathling.views.FhirView;
+import au.csiro.pathling.views.ViewDefinitionGson;
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SaveMode;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.types.StructType;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.BooleanType;
+import org.hl7.fhir.r4.model.IntegerType;
+import org.hl7.fhir.r4.model.Library;
+import org.hl7.fhir.r4.model.Parameters.ParametersParameterComponent;
+import org.hl7.fhir.r4.model.Type;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Orchestrates the execution of the {@code $sqlquery-run} operation. Parses the SQLQuery Library
+ * resource, resolves ViewDefinition dependencies, validates and executes the SQL query, and streams
+ * results in the requested format.
+ */
+@Slf4j
+@Component
+public class SqlQueryExecutionHelper {
+
+  @Nonnull private final SparkSession sparkSession;
+
+  @Nonnull private final QueryableDataSource deltaLake;
+
+  @Nonnull private final FhirContext fhirContext;
+
+  @Nonnull private final ServerConfiguration serverConfiguration;
+
+  @Nonnull private final SqlQueryLibraryParser libraryParser;
+
+  @Nonnull private final SqlValidator sqlValidator;
+
+  @Nonnull private final ViewRegistrationService viewRegistrationService;
+
+  @Nonnull private final ReadExecutor readExecutor;
+
+  @Nonnull private final Gson gson;
+
+  @Nonnull private final ResultStreamingHelper streamingHelper;
+
+  /**
+   * Constructs a new SqlQueryExecutionHelper.
+   *
+   * @param sparkSession the Spark session
+   * @param deltaLake the queryable data source
+   * @param fhirContext the FHIR context
+   * @param serverConfiguration the server configuration
+   * @param libraryParser the SQLQuery Library parser
+   * @param sqlValidator the SQL validator
+   * @param viewRegistrationService the view registration service
+   * @param readExecutor the read executor for reading stored resources
+   */
+  @SuppressWarnings("java:S107")
+  @Autowired
+  public SqlQueryExecutionHelper(
+      @Nonnull final SparkSession sparkSession,
+      @Nonnull final QueryableDataSource deltaLake,
+      @Nonnull final FhirContext fhirContext,
+      @Nonnull final ServerConfiguration serverConfiguration,
+      @Nonnull final SqlQueryLibraryParser libraryParser,
+      @Nonnull final SqlValidator sqlValidator,
+      @Nonnull final ViewRegistrationService viewRegistrationService,
+      @Nonnull final ReadExecutor readExecutor) {
+    this.sparkSession = sparkSession;
+    this.deltaLake = deltaLake;
+    this.fhirContext = fhirContext;
+    this.serverConfiguration = serverConfiguration;
+    this.libraryParser = libraryParser;
+    this.sqlValidator = sqlValidator;
+    this.viewRegistrationService = viewRegistrationService;
+    this.readExecutor = readExecutor;
+    this.gson = ViewDefinitionGson.create();
+    this.streamingHelper = new ResultStreamingHelper(gson);
+  }
+
+  /**
+   * Executes a SQL query from a Library resource and streams results to the HTTP response.
+   *
+   * @param libraryResource the Library resource containing the SQLQuery
+   * @param format the output format, overrides Accept header if provided
+   * @param acceptHeader the HTTP Accept header value, used as fallback if format is not provided
+   * @param includeHeader whether to include a header row in CSV output
+   * @param limit the maximum number of rows to return
+   * @param parameterValues runtime parameter values to bind to the SQL query
+   * @param response the HTTP response for streaming output
+   */
+  @SuppressWarnings("java:S107")
+  public void executeSqlQuery(
+      @Nonnull final IBaseResource libraryResource,
+      @Nullable final String format,
+      @Nullable final String acceptHeader,
+      @Nullable final BooleanType includeHeader,
+      @Nullable final IntegerType limit,
+      @Nullable final List<ParametersParameterComponent> parameterValues,
+      @Nullable final HttpServletResponse response) {
+
+    if (response == null) {
+      throw new InvalidRequestException("HTTP response is required for this operation");
+    }
+
+    // Parse the Library resource as a SQLQuery.
+    final Library library = castToLibrary(libraryResource);
+    final ParsedSqlQuery parsedQuery = libraryParser.parse(library);
+
+    // Resolve ViewDefinitions and check authorization.
+    final Map<String, FhirView> resolvedViews = resolveViewDefinitions(parsedQuery);
+
+    // Validate the SQL structure before execution.
+    sqlValidator.validate(parsedQuery.getSql());
+
+    // Determine output format.
+    final SqlQueryOutputFormat outputFormat =
+        (format != null && !format.isBlank())
+            ? SqlQueryOutputFormat.fromString(format)
+            : SqlQueryOutputFormat.fromAcceptHeader(acceptHeader);
+    final boolean shouldIncludeHeader = includeHeader == null || includeHeader.booleanValue();
+
+    // Register views and execute the query.
+    Map<String, String> registeredViews = Map.of();
+    try {
+      registeredViews = viewRegistrationService.registerViews(resolvedViews, deltaLake);
+
+      // Rewrite SQL to use prefixed temp view names.
+      final String rewrittenSql =
+          viewRegistrationService.rewriteSql(parsedQuery.getSql(), registeredViews);
+
+      // Build parameter map for Spark parameterized queries.
+      final Map<String, String> parameterMap = buildParameterMap(parameterValues);
+
+      // Execute the SQL query. Use parameterized queries only when parameters are provided.
+      Dataset<Row> result;
+      if (parameterMap.isEmpty()) {
+        result = sparkSession.sql(rewrittenSql);
+      } else {
+        @SuppressWarnings("unchecked")
+        final java.util.Map<String, Object> objectMap =
+            (java.util.Map<String, Object>) (java.util.Map<String, ?>) parameterMap;
+        result = sparkSession.sql(rewrittenSql, objectMap);
+      }
+
+      // Apply limit if specified.
+      if (limit != null && limit.getValue() != null) {
+        result = result.limit(limit.getValue());
+      }
+
+      // Stream results.
+      streamResults(result, outputFormat, shouldIncludeHeader, response);
+
+    } finally {
+      viewRegistrationService.dropViews(registeredViews.values());
+    }
+  }
+
+  /** Casts a FHIR resource to a Library, throwing if it is not a Library. */
+  @Nonnull
+  private Library castToLibrary(@Nonnull final IBaseResource resource) {
+    if (resource instanceof final Library library) {
+      return library;
+    }
+    throw new InvalidRequestException(
+        "Expected a Library resource but received: " + resource.fhirType());
+  }
+
+  /**
+   * Resolves ViewDefinition references from the parsed SQL query. Each relatedArtifact reference is
+   * resolved to a FhirView. Authorization is checked for each referenced resource type.
+   */
+  @Nonnull
+  private Map<String, FhirView> resolveViewDefinitions(@Nonnull final ParsedSqlQuery parsedQuery) {
+    final Map<String, FhirView> resolvedViews = new LinkedHashMap<>();
+
+    for (final ViewArtifactReference ref : parsedQuery.getViewReferences()) {
+      final String viewDefinitionId = extractViewDefinitionId(ref.getCanonicalUrl());
+      final IBaseResource viewResource;
+      try {
+        viewResource = readExecutor.read("ViewDefinition", viewDefinitionId);
+      } catch (final Exception e) {
+        throw new InvalidRequestException(
+            "Failed to resolve ViewDefinition for label '"
+                + ref.getLabel()
+                + "' with reference '"
+                + ref.getCanonicalUrl()
+                + "': "
+                + e.getMessage());
+      }
+
+      // Parse the ViewDefinition into a FhirView.
+      final FhirView view = parseViewDefinition(viewResource);
+
+      // Check resource-level authorization for the ViewDefinition's target resource type.
+      if (serverConfiguration.getAuth().isEnabled()) {
+        SecurityAspect.checkHasAuthority(
+            PathlingAuthority.resourceAccess(AccessType.READ, view.getResource()));
+      }
+
+      resolvedViews.put(ref.getLabel(), view);
+    }
+
+    return resolvedViews;
+  }
+
+  /**
+   * Extracts a ViewDefinition ID from a canonical URL or relative reference. Supports relative
+   * references like "ViewDefinition/my-id" and plain IDs.
+   */
+  @Nonnull
+  private String extractViewDefinitionId(@Nonnull final String canonicalUrl) {
+    // Handle relative references like "ViewDefinition/my-id".
+    if (canonicalUrl.contains("/")) {
+      final String[] parts = canonicalUrl.split("/");
+      return parts[parts.length - 1];
+    }
+    // Plain ID.
+    return canonicalUrl;
+  }
+
+  /** Parses a ViewDefinition resource into a FhirView using JSON round-tripping. */
+  @Nonnull
+  private FhirView parseViewDefinition(@Nonnull final IBaseResource viewResource) {
+    try {
+      final String viewJson = fhirContext.newJsonParser().encodeResourceToString(viewResource);
+      return gson.fromJson(viewJson, FhirView.class);
+    } catch (final JsonSyntaxException e) {
+      throw new InvalidRequestException("Invalid ViewDefinition: " + e.getMessage());
+    }
+  }
+
+  /**
+   * Builds a parameter map from FHIR Parameters components. Parameter values are converted to
+   * string representations for Spark's parameterized query API.
+   */
+  @Nonnull
+  private Map<String, String> buildParameterMap(
+      @Nullable final List<ParametersParameterComponent> parameterValues) {
+    final Map<String, String> params = new HashMap<>();
+    if (parameterValues == null) {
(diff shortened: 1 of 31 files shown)


server/src/main/resources/fhir/sqlquery-run.OperationDefinition.json
```diff
@@ -0,0 +1,83 @@
...
+        {
+          "name": "value",
+          "use": "in",
+          "min": 1,
+          "max": "1",
+          "documentation": "The value to bind to the parameter.",
+          "type": "string"
+        }
+      ]
+    }
```
johngrimes, line 81: I think the shape of the parameter input here might not quite match what the spec landed on. The current SQL on FHIR v2 build seems to use a single `parameters` input of type `Parameters` (`0..1`), aligned with the `Measure/$evaluate-measure` envelope in core FHIR. What's here is a repeating `parameter` with a string-typed `value` part, which appears to lose the FHIR type info.

If you're up for switching it over, it should be reasonably mechanical:

1. Could you please drop the repeating `parameter` from the OperationDefinition and add `parameters` as a single `Parameters`-typed input?
2. In the providers, please take it as `@OperationParam(name = "parameters") Parameters parameters` and walk `parameters.getParameter()` for each `name` / `value[x]` pair.
3. While you're in there, please have `buildParameterMap` use the declared `Library.parameter.type` to convert each `value[x]` to a real Java type (Integer, BigDecimal, LocalDate, etc.) rather than calling `primitiveValue()` - that fix folds nicely into the same change.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryOutputFormat.java
```diff
@@ -0,0 +1,140 @@
...
+  NDJSON("ndjson", "application/x-ndjson"),
+
+  /** Comma-separated values format. */
+  CSV("csv", "text/csv"),
+
+  /** JSON format — single document containing an array of objects. */
+  JSON("json", "application/json"),
+
+  /** Apache Parquet columnar format. */
+  PARQUET("parquet", "application/vnd.apache.parquet");
```
johngrimes, line 44: I noticed `_format=fhir` isn't in this enum, which I think is the format that makes the operation feel FHIR-native. The spec describes it as returning a `Parameters` resource - one repeating `row` parameter per result row, each column emitted as a part using the `value[x]` matching the SQL column type, and SQL NULL represented by omitting the part.

If you'd be open to adding it:

- Please add a `FHIR` value to this enum. For wire encoding, I'd suggest letting HAPI's standard content negotiation pick JSON or XML based on the request's `Accept` header (the rest of the server already does this via `setDefaultResponseEncoding(EncodingEnum.JSON)` in `FhirServer.java:287`), so XML support comes "for free" without a separate code path.
- Could you please add a streaming writer that emits the `Parameters` resource row-by-row (to keep memory bounded), using the Spark→FHIR type mapping the spec lays out (BOOLEAN→valueBoolean, INT→valueInteger, TIMESTAMP WITH TIME ZONE→valueInstant, and so on)? Unsupported column types (INTERVAL, ARRAY, XML) would bail out with 422 per the spec.
- And please update the `_format` documentation in `OperationDefinition.json` to mention `fhir` as a supported value.

Spec reference: <https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/OperationDefinition-SQLQueryRun.html>.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryRunProvider.java
```diff
@@ -0,0 +1,90 @@
...
+  @OperationAccess("sqlquery-run")
+  public void run(
+      @Nonnull @OperationParam(name = "queryResource") final IBaseResource queryResource,
+      @Nullable @OperationParam(name = "_format") final String format,
+      @Nullable @OperationParam(name = "header") final BooleanType includeHeader,
+      @Nullable @OperationParam(name = "_limit") final IntegerType limit,
+      @Nullable @OperationParam(name = "parameter")
+          final List<ParametersParameterComponent> parameterValues,
+      @Nonnull final ServletRequestDetails requestDetails,
+      @Nullable final HttpServletResponse response) {
```
johngrimes, line 83: I noticed `queryReference` is in the OperationDefinition (lines 25-32) but neither this method nor `runTypeLevel` seem to bind it - so I think clients reading the CapabilityStatement might assume it's supported when it isn't. The good news is Library already goes through Pathling's generic CRUD pipeline (it's not listed in `ResourceTypes.UNSUPPORTED_RESOURCES`), so resolution shouldn't need any new infrastructure.

If you're up for filling this in, I think a few things on the same change would do it:

- Could you please take `@OperationParam(name = "queryReference") Reference queryReference` alongside `queryResource` on both providers, and mark them both `@Nullable`? Today `queryResource` is `@Nonnull` against an OperationDefinition that says `min: 0`, which I think would NPE inside `castToLibrary` as soon as a request omits it. HAPI doesn't enforce JSR-305 annotations as far as I can tell, so the `@Nonnull` may be giving false confidence here.
- Please validate "exactly one of `queryResource` / `queryReference`" at the top of `executeSqlQuery`, returning 400 with an `OperationOutcome` if neither or both are provided.
- Please resolve the literal reference in `Reference.reference` in both forms defined by [FHIR References](https://hl7.org/fhir/references.html#Reference):
  - Relative literal references (`Library/abc`): pull the id and call `readExecutor.read("Library", id)` - same path `runById` already uses.
  - Canonical references (`[url]` or `[url]|[version]` matching `Library.url`): search the data source for the matching Library, and if there's a `|version` suffix, narrow on `Library.version` too. If multiple match, prefer an exact `url|version`, otherwise the latest active version.
- Please return 404 when the reference doesn't resolve, matching the `ResourceNotFoundException` behaviour `runById` already has.
- And if you have time, a couple of examples in `examples/sqlquery-run-examples.md` (one relative, one canonical) would be great.

server/src/main/resources/fhir/sqlquery-run.OperationDefinition.json
```diff
@@ -0,0 +1,83 @@
...
+      "type": "Reference"
+    },
+    {
+      "name": "_format",
+      "use": "in",
+      "min": 0,
+      "max": "1",
+      "documentation": "The output format for the query results. Supported values are 'ndjson' (default), 'csv', 'json', and 'parquet'.",
+      "type": "code"
+    },
```
johngrimes, line 40: Just a heads-up: I noticed the spec has `_format` at `min: 1`, while the PR uses `min: 0` with an `ndjson` default and Accept-header fallback. I think the PR's design is the more usable one and would be consistent across the three SQL on FHIR operations, so I've opened <https://github.com/FHIR/sql-on-fhir-v2/issues/345> upstream to suggest the spec adopt that pattern. If it lands, nothing to do here. If it doesn't, this OperationDefinition would need to flip to `min: 1` and the providers would need to reject requests that omit `_format`. Let's keep this the way it is and track the upstream change.

server/src/main/resources/fhir/sqlquery-run.OperationDefinition.json
```diff
@@ -0,0 +1,83 @@
...
+      "type": "boolean"
+    },
+    {
+      "name": "_limit",
+      "use": "in",
+      "min": 0,
+      "max": "1",
+      "documentation": "Maximum number of rows to return in the result.",
+      "type": "integer"
+    },
```
johngrimes, line 56: `_limit` isn't in the `$sqlquery-run` spec today, but `$viewdefinition-run` defines it (`operations.fsh:312-320`) and your signature matches it exactly - so I think this looks more like a spec inconsistency than anything added here. I've opened <https://github.com/FHIR/sql-on-fhir-v2/issues/346> upstream asking for `_limit` to be added to `$sqlquery-run` too. If that lands, the parameter is already shaped correctly and nothing needs to change here.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryLibraryParser.java
```diff
@@ -0,0 +1,143 @@
...
+            "Each relatedArtifact in the SQLQuery Library must have a label");
+      }
+      if (resource == null || resource.isBlank()) {
+        throw new InvalidRequestException(
+            "Each relatedArtifact in the SQLQuery Library must have a resource reference");
+      }
+      references.add(new ViewArtifactReference(label, resource));
+    }
+    return references;
+  }
```
johngrimes, line 117: I noticed the parser is only checking that label and resource are non-blank - I don't think any of the [SQLQuery profile](https://build.fhir.org/ig/FHIR/sql-on-fhir-v2/StructureDefinition-SQLQuery.html) invariants are actually being enforced yet. So malformed inputs reach Spark and either fail with confusing knock-on errors or, in the label case, end up registering arbitrary strings as temp view names.

Could you please add the profile checks here at parse time, returning 400 with a clear message?

- Please require `Library.type` to have a coding with `system = https://sql-on-fhir.org/ig/CodeSystem/LibraryTypesCodes` and `code = sql-query`.
- Please require `relatedArtifact.type` to be `depends-on` (rejecting `cite-as`, `documentation`, etc.).
- Please enforce `relatedArtifact.label` against the regex `^[A-Za-z][A-Za-z0-9_]*$` - this is a profile invariant and also removes the identifier-injection risk on the temp-view registration side.
- Please require `parameter.use` to be `in` (the profile fixes this; `out` doesn't mean anything on input).

If you have time, a handful of unit tests in `SqlQueryLibraryParserTest` covering each rejection case would be great.

server/src/main/java/au/csiro/pathling/operations/sqlquery/ViewRegistrationService.java
```diff
@@ -0,0 +1,124 @@
...
+        // Drop any views that were already registered before failing.
+        dropViews(registeredViewNames);
+        throw new InvalidRequestException(
+            "Failed to execute ViewDefinition for label '" + label + "': " + e.getMessage());
+      }
+
+      result.createOrReplaceTempView(label);
+      registeredViewNames.add(label);
+      log.info("Registered temporary view '{}' for resource type '{}'", label, view.getResource());
+    }
```
johngrimes, line 104: I think there might be a concurrency issue here that's worth a closer look. `createOrReplaceTempView(label)` writes into the session-global catalog, and Pathling shares one `SparkSession` across the whole server - "temp" really only means "lifetime tied to the session", not per-request. So I think two `$sqlquery-run` requests that pick the same label (which seems likely given how naturally names like `patients` or `observations` come up) might clobber each other: request B replaces A's plan, A's `finally` drops the view while B is mid-flight, that kind of thing.

Looking at the history, I think the original commit (`ad09aaa6`) actually had this covered - it prefixed view names with a UUID and rewrote the SQL accordingly. Commit `815146f5` ("adding samples to the mix") seems to have taken the prefixing out without replacing it, which I think is the regression.

For the fix, the good news is that you don't need to mint a new UUID - HAPI is already generating a per-request id and stashing it on `RequestDetails`. `RestfulServer.java:1037-1038` calls `setRequestId` for every request (16 alphanumeric chars from `RandomStringUtils`, or whatever's in `X-Request-ID`), and `DiagnosticContext.java:73` is already reading it for Sentry. So:

- Could you please pull `requestDetails.getRequestId()` out in `executeSqlQuery` and pass it through to `registerViews`?
- Please register views as `sqlquery_<requestId>_<label>` and rewrite the SQL to substitute each `label` (token-aware, not naive `String.replace`) before handing it to Spark - the old commit already has the rewriting code, so it should be lift-able.

Small bonus: the request id ends up in access logs and error messages too, so a problematic SQL query becomes easier to correlate end-to-end.

If you have time, please add a concurrency test - two parallel requests, both registering a view called `patients` with different underlying data, asserting each sees its own. It should fail against the current code and pass after the fix. Happy to chat if I'm misreading any of this.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryExecutionHelper.java
```diff
@@ -0,0 +1,428 @@
...
+        }
+      }
+
+      if (name != null && value != null) {
+        params.put(name, value);
+      }
+    }
+
+    return params;
+  }
```
johngrimes, line 322: I noticed calling `valueType.primitiveValue()` here collapses every parameter to a Java `String`, and combined with the unchecked double-cast at lines 197-200 I think this means Spark sees every parameter as a string literal regardless of the declared FHIR type. So I think `WHERE age > :min_age` with `valueInteger=42` becomes `WHERE age > '42'` after binding, which leaves things at the mercy of Spark's coercion rules - it'll error or silently produce wrong results depending on the column type. Dates, decimals, booleans all break the same way as far as I can tell.

In case it helps: Spark's parameter binding is actually properly typed, you just need to give it real values. `SparkSession.sql(String, Map<String, Object>)` (`SparkSession.scala:511`, since 3.4) takes `Any` values and runs each through `lit(arg).expr` → `Literal.create(value)`, which infers the SQL type from the Java class. `BindParameters` then substitutes the typed literal into the plan during analysis. So passing an `Integer` gets you an integer literal; passing a `LocalDate` gets you a date literal. No string interpolation anywhere - SQL injection is structurally impossible.

Could you please use the declared `Library.parameter.type` to convert each `value[x]` to the right Java type before adding it to the map? Something roughly like:

| FHIR type                                                               | Java type              |
| ----------------------------------------------------------------------- | ---------------------- |
| `boolean`                                                               | `java.lang.Boolean`    |
| `integer` / `unsignedInt` / `positiveInt`                               | `java.lang.Integer`    |
| `decimal`                                                               | `java.math.BigDecimal` |
| `string` / `code` / `id` / `uri` / `url` / `canonical` / `oid` / `uuid` | `java.lang.String`     |
| `date`                                                                  | `java.time.LocalDate`  |
| `dateTime` / `instant`                                                  | `java.time.Instant`    |
| `time`                                                                  | `java.time.LocalTime`  |
| `base64Binary`                                                          | `byte[]`               |

While you're in there, a couple of related cleanups worth folding in if you're up for it:

- Please build the map as `LinkedHashMap<String, Object>` from the start with typed values - I think the `@SuppressWarnings("unchecked")` double-cast at 197-200 then goes away on its own.
- Please return 400 if a runtime value doesn't match the declared type (e.g. `valueString` supplied for an `integer` parameter).
- Please return 400 if a runtime parameter name isn't declared in `Library.parameter` - today I think those silently pass through.
- This one goes well with the switch to a `parameters` Parameters input - that's where `value[x]` arrives properly typed, so doing both together would be one pass through the data.

If you have time, a few test cases in a new `SqlQueryExecutionHelperTest` would lock the behaviour in - integer parameter producing an integer comparison, date round-tripping through `WHERE birth_date >= :since`, type mismatch returning 400, undeclared name returning 400.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlValidator.java
```diff
@@ -0,0 +1,461 @@
...
+          "TryEval",
+          "AssertTrue",
+          // Utility.
+          "Empty2Null",
+          "TypeOf",
+          // Bitwise operations.
+          "BitwiseAnd",
+          "BitwiseOr",
+          "BitwiseXor",
+          "BitwiseNot");
```
johngrimes, line 294: Nothing here is breaking the operation - the validator looks like a solid first pass. Since this is the security boundary for an endpoint that runs user-supplied SQL though, I wanted to flag a few places that I think could be tightened a bit, if you're up for it.

1. **Could you please consider matching on full class names rather than simple names?** `validatePlanNode` and `validateExpression` (`:383-389`, `:399-425`) use `getClass().getSimpleName()`. I think a future or third-party Spark plugin shipping a `Project` or `Filter` in a different package would slip through silently. `getClass().getName()` against fully-qualified names would close that gap. Where the allow-list naturally sits at a hierarchy root (`BinaryArithmetic`, `BinaryComparison`, etc.), `instanceof` against the base class might be even better - shorter list and more robust across Spark upgrades.

2. **Please consider adding an allow-list integrity test.** A parameterised test asserting that every entry in `ALLOWED_PLAN_NODES`, `ALLOWED_EXPRESSION_NAMES`, etc. actually resolves to a class on the classpath would catch typos and Spark-rename drift at build time, rather than at runtime when a real user query suddenly gets rejected.

3. **Please consider validating the analysed plan, not just the unresolved one.** `validate()` currently runs against `sqlParser().parsePlan(sql)` (`:329`). I think some unsafe constructs only show up after analysis - `HiveSimpleUDF` and `HiveGenericUDF` are inserted by the analyser when a Hive function gets resolved, so they aren't in the unresolved tree at all. They're listed in `REJECTED_EXPRESSION_NAMES` but I think the check never fires because the expression isn't there yet. Could you either run validation again after `Dataset.queryExecution().analyzed`, or move it post-analyse entirely? This is the most architecturally significant of the five.

4. **Could you please pull the UDF allow-list from the registrars?** `ALLOWED_UDF_NAMES` (`:133-158`) is currently a hand-copy of names from `TerminologyUdfRegistrar` and `PathlingUdfRegistrar`, which I think will be easy to drift out of sync. Either having the registrars expose their names, or having `SqlValidator` discover them via the Spark function registry filtered by the Pathling prefix, would give one source of truth.

5. **Optional, but nice to have.** If you have time, a regression test corpus of known-bad queries (`reflect()`, `java_method()`, INSERT, UPDATE, DROP, CREATE TABLE, hive UDFs, `LOAD DATA INPATH`, `MSCK REPAIR`, `SET`, etc.) would lock current rejection behaviour in place and double as documentation of the security model.

(1), (2), and (4) feel like small mechanical changes; (3) is the bigger one. Let me know if you have any questions.

server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryExecutionHelper.java
```diff
@@ -0,0 +1,428 @@
...
+        result = result.limit(limit.getValue());
+      }
+
+      // Stream results.
+      streamResults(result, outputFormat, shouldIncludeHeader, response);
+
+    } finally {
+      viewRegistrationService.dropViews(registeredViewNames);
+    }
+  }
```
johngrimes, line 214: I would like to suggest a refactoring here. By my count, `executeSqlQuery` is currently quite a few responsibilities (cast/parse Library, resolve ViewDefinitions, per-view auth, SQL validation, format selection, parameter map building, temp-view registration, SQL execution, limit, format-branched streaming, Parquet temp file lifecycle, temp-view cleanup). The two `@SuppressWarnings("java:S107")` annotations on the constructor and method are SonarQube already nudging in this direction.

  The reason I'm flagging it: other changes in the review lands in this same method. If they go in here in place, I worry `executeSqlQuery` will get steadily harder to follow. If you split first, each fix slots into a single small class instead.

  Could you please consider a split along these lines?

  | Class                                                     | Responsibility                                                                                                                                                                  | Deps                                                      |
  | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------- |
  | `SqlQueryRequestParser` (new)                             | Parses raw HTTP inputs (`queryResource` / `queryReference`, `parameters`, `_format`, `header`, `_limit`) into a validated `SqlQueryRequest` value object. No Spark dependency.  | `SqlQueryLibraryParser`, `ReadExecutor`                   |
  | `ViewResolver` (new, or extend `ViewRegistrationService`) | Turns each `ViewArtifactReference` into a resolved `FhirView` and does the per-resource-type auth check. No Spark dependency.                                                   | `ReadExecutor`, `ServerConfiguration`, `FhirContext`      |
  | `SqlQueryExecutor` (new)                                  | The only piece that touches Spark. Registers views with the request-id prefix, rewrites the SQL, runs it, applies limit, returns the `Dataset<Row>`. Owns the §2.1 namespacing. | `SparkSession`, `ViewRegistrationService`, `SqlValidator` |
  | `SqlQueryResultStreamer` (new)                            | Format dispatch: NDJSON/CSV/JSON via the existing `ResultStreamingHelper`, plus Parquet and (for §1.2) FHIR.                                                                    | `ResultStreamingHelper`                                   |
  | `SqlQueryExecutionHelper` (slimmed)                       | Orchestrator. Chains the four stages. ~30 lines.                                                                                                                                | The four above                                            |

  Order-wise, if you do go this route, I'd gently suggest landing the refactor first (either as the first commit on the branch or as a separate prep PR), then doing the substantive fixes in the new structure. Otherwise I think you'd be loading more onto a method that's already at the limit. Totally happy to discuss if you'd prefer to keep things as-is or have a different split in mind.

server/src/main/java/au/csiro/pathling/operations/update/UpdateExecutor.java
```diff
@@ -129,6 +134,28 @@ public void merge(
...
+        // without inserting any data. The MERGE that follows then sees compatible schemas and
+        // succeeds.
+        updates
+            .limit(0)
+            .write()
+            .format("delta")
+            .mode(SaveMode.Append)
+            .option("mergeSchema", "true")
+            .save(tablePath);
+      }
```
johngrimes, line 157: RE: 0d24b7191c and 4f3fc61425

These two commits look like a useful piece of work in their own right - they seem to add an opt-in `schemaAutoMerge` flag that lets writes succeed against Delta tables whose schema lags the current FHIR encoder, specifically the case where new fields appear inside nested array-of-struct elements (new value-types inside `extension[].value[x]`, that kind of thing), which Delta's MERGE can't handle even with `autoMerge.enabled=true`. The workaround of a zero-row append with `mergeSchema=true` to widen the target schema first, then run the MERGE, is clever.

I think this is genuinely valuable for anyone upgrading a long-lived Pathling warehouse across encoder-schema changes. The thing is, it's a write-path change with no functional connection to `$sqlquery-run`. My best guess is you needed writes to keep working against an existing warehouse to be able to test the new SQL operation end-to-end, so this workaround naturally landed on the same branch.

Could you please split this into its own PR with its own issue? It touches only `UpdateExecutor.java`, `StorageConfiguration.java`, `application.yml`, and `ViewDefinitionCreateTest.java`, so I think it should lift cleanly.

One small thing worth flagging for that separate PR: every merge against a flag-enabled table now does two Delta transactions (the warmup write, then the actual merge), so the table version increments twice per logical update even when no schema change was needed. A quick "do schemas already match?" check would skip the warmup in the common case.

johngrimes (changes requested):
Hi @jkiddo,

Thanks so much for your work on this!

This is a large change, and I've picked up on a few areas that still need some work.

If you like, we could relocate this branch to the main repo and give you committer access. This would allow us to work together on the branch, rather than you making all the changes yourself.

Let me know what you think.

jkiddo:
> Hi @jkiddo,
> 
> Thanks so much for your work on this!
> 
> This is a large change, and I've picked up on a few areas that still need some work.
> 
> If you like, we could relocate this branch to the main repo and give you committer access. This would allow us to work together on the branch, rather than you making all the changes yourself.
> 
> Let me know what you think.

yea - that would make good sense. I'm up for that.

server/src/main/java/au/csiro/pathling/operations/update/UpdateExecutor.java
jkiddo: totally ... the branch got a little overheated as it went from a more clean feature extension branch - to also accomodate for some issues we found 😆 - I'll move this to another feature branch.

jkiddo: @johngrimes - I've made the branch here: https://github.com/aehrc/pathling/tree/feature/schema-evolution feel free to rename it if there is a certain convention that should be adhered to. That branch has only the schemaMerge feature set.

jkiddo:
I'll move the contents of this branch as well to this repo

Issue #2378, opened 2025-06-06
Support all subsetting functions


johngrimes:
Spec link: https://hl7.org/fhirpath/#subsetting

This includes the following functions:

- single
- tail
- take
- skip
- last
- exclude
- intersect

PR #2587, opened 2026-04-11
feat: Support FHIRPath union() and combine() combining functions

### LINKED ISSUE #2384: Support combining functions
Spec link: https://hl7.org/fhirpath/#combining

This includes the `union` and `combine` functions.

### LINKED ISSUE #2588: bug: Quantity canonicalisation produces heterogeneous struct shapes, breaking repeat() over mixed-unit collections
## Summary

When a FHIRPath collection contains Quantity values with different structural shapes (for example, an indefinite calendar duration alongside a UCUM quantity, or UCUM quantities with different canonical units that normalise differently), applying `repeat($this)` over that collection fails. The failure originates in Pathling's Quantity canonicalisation: structurally-different Quantity literals produce structs whose subfields differ in **nullability**, and `repeat()` subsequently casts those struct elements to `VARIANT` for deduplication, which rejects the heterogeneous shape.

This bug is **independent of `combine()`** — `combine()` is simply the cleanest way to build a collection that exposes it, because it concatenates without deduplicating away the mismatched structs. The same underlying canonicalisation shape mismatch is also what produces the `wontfix` "Indefinite calendar duration union deduplication" exclusion (`(1 year | 12 months)`) in `fhirpath/src/test/resources/fhirpath-js/config.yaml`.

## Reproduction

Both expressions fail in Pathling and are currently excluded from the `fhirpath-js` reference test suite (`config.yaml:179-192`):

```fhirpath
(1 year).combine(12 months).repeat(\$this)
(3 'min').combine(180 seconds).repeat(\$this)
```

The first case mixes an indefinite calendar duration (year) with a definite one (month); the second mixes two UCUM units that canonicalise to different base representations.

## Root cause (suspected)

Pathling's Quantity encoding uses a struct with fields for the original value/unit/system/code plus canonicalised value/unit fields used for equality and comparison. Indefinite calendar durations (year, month) leave the canonical subfields **null**, whereas UCUM quantities populate them. When two literals of these different classes end up in the same array, the resulting Spark array contains structs whose subfield nullabilities are inconsistent.

`repeat()`'s deduplication path casts the element struct to `VARIANT`, and Spark's `VARIANT` casting is strict about schema homogeneity within an array, so the cast fails.

## Expected behaviour

`repeat(\$this)` over a heterogeneous Quantity collection should return the collection (each element is its own fixed point under `\$this`). It should not raise a canonicalisation/cast error.

## Suggested fix directions

1. **Normalise Quantity struct shape at encoding time** — emit a consistent schema where every subfield is always present (possibly with null values) regardless of whether the literal is a calendar duration or a UCUM quantity.
2. **Avoid `VARIANT` casting in `repeat()`** for struct-typed collections — use a deduplication strategy that honours the collection's comparator instead, mirroring the approach already used by `UnionOperator` / `CombiningLogic.dedupeArray` for types with custom equality.

Option 1 would also unblock the related `wontfix` exclusion on indefinite calendar duration union deduplication.

## Impact

- Currently blocks `repeat()` / deduplication-based operations over any collection that mixes structurally-different Quantity literals.
- Surfaces via `combine()` (new in #2384) but is a latent issue in the Quantity encoder, not in the combining functions.

## References

- Discovered while implementing #2384 (FHIRPath `union()` / `combine()` functions), PR #2587.
- Exclusion: `fhirpath/src/test/resources/fhirpath-js/config.yaml:179-192`.
- Related `wontfix` exclusion: "Indefinite calendar duration union deduplication" in the same file.

piotrszul:
## Summary

- Closes #2384 by implementing the FHIRPath `union(other)` and `combine(other)` functions from §6.5 Combining.
- Closes #2588 by fixing VOID-typed fields in Quantity literal structs that prevented `to_variant_object()` conversion in `repeat()`/`repeatAll()`.
- Both function forms are parser-desugared into `EvalOperator` ASTs so that `x.union(y)` is strictly equivalent to `x | y` (including inside iteration contexts such as `name.select(use.union(given))`), and `x.combine(y)` reaches a new peer `CombineOperator` that merges without deduplicating.
- Shared `CombiningLogic` helper holds the array-level merge primitives (Decimal normalization, comparator-aware dedup/concat) used by both `UnionOperator` and `CombineOperator`.

## Why parser desugaring

The spec explicitly declares `x.union(y)` to be synonymous with `x | y`, and the worked example `name.select(use.union(given)) ≡ name.select(use | given)` requires the `given` argument to resolve against the current iteration focus. Routing the argument through Pathling's standard function invocation path would evaluate `given` against `%context` (the outer root focus), losing the per-name iteration. Desugaring `x.union(y)` and `x.combine(y)` at parse time into operator-form ASTs gives this equivalence by construction, without refactoring `Composite.apply` / `EvalFunction` / `FunctionParameterResolver`.

The full design rationale lives in `openspec/changes/archive/2026-04-11-add-combining-functions/design.md`.

## Quantity VOID fix (#2588)

`QuantityEncoding.encodeLiteral` used `lit(null)` for absent fields, producing Spark `NullType` (VOID) in the struct schema. `to_variant_object()` rejects VOID fields, breaking `repeat($this)` over any Quantity literal collection. Fixed by casting the struct to `dataType()` after construction and fixing `FlexiDecimalSupport.toLiteral` to return a typed null.

`(3 'min').combine(180 seconds).repeat($this)` now passes. `(1 year).combine(12 months).repeat($this)` is reclassified as wontfix (year-to-month equality semantics, not a VOID type issue).

## Test plan

- [x] `mvn test -pl fhirpath -Dtest=CombiningFunctionsDslTest` — 78 new assertions covering the type matrix, duplicate-preservation, `union() ≡ \|` equivalence across 9 types (Boolean, Integer, Decimal, String, Date, DateTime, Time, Quantity, Coding), iteration-context pin-downs inside `select`, incompatible-type errors, and arity errors
- [x] `mvn test -pl fhirpath -Dtest=CombiningOperatorsDslTest` — 188/188 pass (refactor behaviour-preserving for `\|`)
- [x] `mvn test -pl fhirpath -Dtest=RepeatFunctionDslTest` — 23/23 pass including 2 new #2588 regression tests
- [x] `mvn test -pl fhirpath -Dtest=YamlReferenceImplTest` — 1821 run, 0 failures, 979 skipped; many previously-failing reference tests for `.combine(...)` now pass
- [x] `mvn test -pl fhirpath` — 6388 tests, no new failures

## Documentation

`site/docs/fhirpath/index.md` gains a new "Combining functions" section listing `union()` and `combine()`, and the `|` operator row cross-references the `union()` function.

### DIFF
--- a/encoders/src/main/scala/au/csiro/pathling/sql/types/FlexiDecimalSupport.scala
+++ b/encoders/src/main/scala/au/csiro/pathling/sql/types/FlexiDecimalSupport.scala
@@ -47,7 +47,7 @@ object FlexiDecimalSupport {
         lit(normalizedValue.scale()).as("scale")
       )
     } else {
-      lit(null)
+      lit(null).cast(FlexiDecimal.DATA_TYPE)
     }
   }
 
--- a/fhirpath/src/main/java/au/csiro/pathling/fhirpath/encoding/QuantityEncoding.java
+++ b/fhirpath/src/main/java/au/csiro/pathling/fhirpath/encoding/QuantityEncoding.java
@@ -253,17 +253,22 @@ public class QuantityEncoding {
   public static Column encodeLiteral(@Nonnull final FhirPathQuantity quantity) {
     final BigDecimal value = quantity.getValue();
     @Nullable final Ucum.ValueWithUnit canonical = canonicalOf(quantity);
+    // Cast the struct to the canonical Quantity schema so that fields set to lit(null) carry their
+    // declared types instead of Spark's NullType (VOID). VOID-typed fields prevent the struct from
+    // being converted to VARIANT, which is required by variantTransformTree in
+    // repeat()/repeatAll().
     return toStruct(
-        lit(null),
-        lit(value),
-        lit(value.scale()),
-        lit(null),
-        lit(quantity.getUnitName()),
-        lit(quantity.getSystem()),
-        lit(quantity.getCode()),
-        FlexiDecimalSupport.toLiteral(canonical != null ? canonical.value() : null),
-        lit(canonical != null ? canonical.unit() : null),
-        lit(null));
+            lit(null),
+            lit(value),
+            lit(value.scale()),
+            lit(null),
+            lit(quantity.getUnitName()),
+            lit(quantity.getSystem()),
+            lit(quantity.getCode()),
+            FlexiDecimalSupport.toLiteral(canonical != null ? canonical.value() : null),
+            lit(canonical != null ? canonical.unit() : null),
+            lit(null))
+        .cast(dataType());
   }
 
   /**
--- a/fhirpath/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java
+++ b/fhirpath/src/main/java/au/csiro/pathling/fhirpath/operator/CombineOperator.java
@@ -28,15 +28,8 @@ import org.apache.spark.sql.Column;
  * of order in the resulting collection.
  *
  * <p>Unlike {@link UnionOperator}, {@code combine} does not deduplicate and does not need to
- * consult the collection's equality comparator. Type reconciliation, Decimal normalization, and
- * empty-operand dispatch are inherited from {@link SameTypeBinaryOperator}, and the array-level
- * merge primitive is shared with {@link UnionOperator} via {@link CombiningLogic}.
- *
- * <p>FHIRPath does not define a symbolic operator for {@code combine}. This class is reachable only
- * via the parser's invocation visitor, which desugars {@code x.combine(y)} into an {@code
- * EvalOperator} using this operator as the merge step. This mirrors the desugaring used by {@code
- * x.union(y)} and ensures that {@code combine()} honours the surrounding iteration context in the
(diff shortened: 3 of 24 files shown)


johngrimes:
@piotrszul Regarding the need for "desugaring"...

Have you given any thought to whether this will be required for other functions on the roadmap?

For example: #2378 

Do we have a limitation in our function implementation framework more generally?

piotrszul:
@johngrimes This is a very astute question and the answer it most likely yes (or it depends).

in essence the question is if 

`where(x.func(y))` 

resolves to
a)  `where($this.x.func($this.y))` 
b)  `where($this.x.func(%context.y))`

The specification is not clear on this (except for the union case)
The implementations work differently:
- .NET(firefly) : always (a)
- Aidbox: always(b)
- reference.js   (a) for union, combine, intersect, exclude, subsetOf, aggregate  and (b) for everything else. Possibly the difference in the specification is the type of the argument 'collection' vs 'String', 'Integer' etc).

We currently always do (b) with the exception of union and combine which effectively use (a) because of desugaring.

So yes, depending on what we choose we might need to extend the framework.

PR #2589, opened 2026-04-17
feat: Group evaluation results by context element

### LINKED ISSUE #2586: FHIRPath Lab API should group on each context element
There is an optional field in FHIRPath Lab called "Context Expression".

The way this works for other implementations is that the main expression is executed for each element returned by the context expression.

The Pathling FHIRPath Lab server doesn't currently work this way, it simply returns a single collection.

piotrszul:
## Summary
- **BREAKING**: Replace flat `results`/`traces` on `SingleInstanceEvaluationResult` with a uniform `resultGroups` list of `ResultGroup` objects, each containing a context key, results, and traces
- Implement per-context-element evaluation in `SingleInstanceEvaluator` — the main expression is evaluated independently for each context element with isolated trace collection
- **BREAKING**: Update Python `evaluate_fhirpath` return structure to use `resultGroups` instead of top-level `results`/`traces`
- Add new `ResultGroup` class, `ListTraceCollector.clear()` method, and comprehensive tests for context grouping, trace isolation, empty/scalar context handling

Related to 2586

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,14 +76,23 @@ jobs:
       - name: Run install goal
         env:
           R_KEEP_PKG_SOURCE: yes
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
         run: |
           mvn --batch-mode install \
+          -pl '!site,!benchmark'
+        timeout-minutes: 60
+
+      - name: Run SonarCloud analysis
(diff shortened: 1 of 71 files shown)


piotrszul:
Made obsolete by #2593.
The functionality of #2586 have been implemented entirely on the fhirpath-lab-api side.

PR #2591, opened 2026-04-20
Enable FHIR extensions in FHIRPath Lab API context

### LINKED ISSUE #2590: Enable FHIR extensions support in FHIRPath Lab API
The FHIRPath Lab API constructs its `PathlingContext` without enabling extensions support, which means FHIRPath expressions that traverse the `extension` element on FHIR resources return empty collections.

This limits the usefulness of the API for evaluating expressions against real-world FHIR resources, particularly those conforming to implementation guides such as AU Base or US Core, where extensions are heavily used.

## Change required

Pass `enable_extensions=True` when creating the `PathlingContext` in `fhirpath-lab-api/src/fhirpath_lab_api/app.py`:

```python
ctx = PathlingContext.create(enable_extensions=True)
```

johngrimes:
Closes #2590.

Pass `enable_extensions=True` when constructing the `PathlingContext` in the FHIRPath Lab API so that FHIRPath expressions can traverse the `extension` element on FHIR resources.

### DIFF
--- a/fhirpath-lab-api/src/fhirpath_lab_api/app.py
+++ b/fhirpath-lab-api/src/fhirpath_lab_api/app.py
@@ -67,7 +67,7 @@ def create_app(pathling_context=None) -> Flask:
             from pathling import PathlingContext
 
             logger.info("Initialising PathlingContext...")
-            ctx = PathlingContext.create()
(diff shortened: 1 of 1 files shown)


piotrszul (approved):
Looks good

PR #2593, opened 2026-04-24
FHIRPath Lab context expression grouping and trace support

### LINKED ISSUE #2585: Include trace entries in FHIRPath Lab API response
Depends on #2580.

Update the FHIRPath Lab API to include trace entries (from the `trace()` function implemented in #2580) in the FHIR Parameters response, so they can be displayed in the FHIRPath Lab UI.

### fhirpathlab-api

Update `build_response_parameters()` to include trace entries as additional parameter parts in the FHIR Parameters response, e.g.:

```json
{
  "name": "trace",
  "part": [
    {"name": "name", "valueString": "unmatched"},
    {"name": "value", "valueString": "[id1, id2]"}
  ]
}
```

One `trace` parameter per `trace()` call encountered during evaluation.

### LINKED ISSUE #2586: FHIRPath Lab API should group on each context element

piotrszul:
## Summary
- Group FHIRPath Lab results by context element so the main expression is evaluated per context item (#2586).
- Include `trace()` entries in the FHIR Parameters response so they can be surfaced in the FHIRPath Lab UI (#2585).
- Bump `fhirpath-lab-api` to pathling 9.6.0.

Closes #2585, closes #2586.

## Test plan
- [x] `uv run pytest` in `fhirpath-lab-api`
- [x] Smoke test a FHIRPath Lab request with a Context Expression and verify per-context grouping
- [x] Smoke test an expression using `trace()` (unary and binary) and verify trace parts in the response

### DIFF
--- a/fhirpath-lab-api/README.md
+++ b/fhirpath-lab-api/README.md
@@ -118,6 +118,76 @@ Response (FHIR Parameters):
 }
 ```
 
+#### Trace entries
+
+Each invocation of the FHIRPath `trace()` function inside the expression
+contributes an additional part inside the `result` part. The trace part is
+named `trace`, its `valueString` is the trace label, and its `part` list
+contains the typed values that flowed through the trace. Trace parts appear
+after the typed result values within the same `result` part.
+
+For an expression such as `name.given.trace('all-given').first()`:
+
+```json
+{
+    "name": "result",
+    "part": [
+        { "name": "string", "valueString": "John" },
+        {
+            "name": "trace",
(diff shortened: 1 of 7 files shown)


johngrimes (approved):
Looks good to me, thanks @piotrszul!

PR #2616, opened 2026-05-19
fix: Resolve intermittent test failures caused by shared state leaks

### LINKED ISSUE #2615: Server tests fail intermittently due to spark.stop() destroying shared SparkContext
## Description

Several server tests fail intermittently with Spark session errors when the full test suite is run. The failures are non-deterministic — different tests fail on different runs — because they depend on which tests happen to execute first.

## Failing tests

- `ViewDefinitionSearchTest` — 9 errors
- `ViewDefinitionCreateTest` — 6 errors

## Error extract

```
[ERROR] Tests run: 9, Failures: 0, Errors: 9, Skipped: 0, Time elapsed: 2.911 s <<< FAILURE! -- in au.csiro.pathling.operations.view.ViewDefinitionSearchTest
[ERROR] au.csiro.pathling.operations.view.ViewDefinitionSearchTest.searchViewDefinitionsWithNameFilter -- Time elapsed: 0.320 s <<< ERROR!
org.apache.spark.SparkException: [INTERNAL_ERROR] The SparkPlan.session() method is called but the associated SparkSession is null.

[ERROR] Tests run: 6, Failures: 0, Errors: 6, Skipped: 0, Time elapsed: 0.296 s <<< FAILURE! -- in au.csiro.pathling.operations.view.ViewDefinitionCreateTest
[ERROR] au.csiro.pathling.operations.view.ViewDefinitionCreateTest.createViewDefinitionGeneratesNewUuid -- Time elapsed: 0.296 s <<< ERROR!
org.apache.spark.SparkException: [INTERNAL_ERROR] No active or default Spark session found SQLSTATE: XX000
	at org.apache.spark.SparkException$.internalError(SparkException.scala:92)
	at org.apache.spark.SparkException$.internalError(SparkException.scala:96)
	at org.apache.spark.sql.SparkSessionCompanion.$anonfun$active$2(SparkSession.scala:1031)
	at scala.Option.getOrElse(Option.scala:201)
	at org.apache.spark.sql.SparkSessionCompanion.$anonfun$active$1(SparkSession.scala:1031)
	at scala.Option.getOrElse(Option.scala:201)
	at org.apache.spark.sql.SparkSessionCompanion.active(SparkSession.scala:1030)
	at org.apache.spark.sql.delta.DeltaTableUtils$.findDeltaTableRoot(DeltaTable.scala:206)
	at org.apache.spark.sql.delta.DeltaTableUtils$.findDeltaTableRoot(DeltaTable.scala:198)
	at io.delta.tables.DeltaTable$.isDeltaTable(DeltaTable.scala:1283)
	at io.delta.tables.DeltaTable.isDeltaTable(DeltaTable.scala)
	at au.csiro.pathling.operations.update.UpdateExecutor.deltaTableExists(UpdateExecutor.java:243)
	at au.csiro.pathling.operations.update.UpdateExecutor.merge(UpdateExecutor.java:145)
	at au.csiro.pathling.operations.update.UpdateExecutor.merge(UpdateExecutor.java:121)
	at au.csiro.pathling.operations.create.CreateProvider.create(CreateProvider.java:111)
	at au.csiro.pathling.operations.view.ViewDefinitionCreateTest.createViewDefinitionGeneratesNewUuid(ViewDefinitionCreateTest.java:119)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
```

## Root cause

Three standalone tests created their own `SparkSession` via `SparkSession.builder().getOrCreate()` and called `spark.stop()` in an `@AfterAll` lifecycle method:

- `SqlQueryResultStreamerTest`
- `ViewRegistrationServiceTest`
- `LibraryReferenceResolverTest.CanonicalReferences`

Because `getOrCreate()` returns the JVM-wide singleton `SparkSession`, calling `spark.stop()` on any instance destroys the shared `SparkContext` for the entire JVM process. Any test that runs afterward and tries to use Spark fails with "no active or default Spark session found" or a null `SparkPlan.session()`.

The failures were non-deterministic because they only occurred when one of the three offending tests happened to run before `ViewDefinitionSearchTest` or `ViewDefinitionCreateTest` within the same JVM.

## Fix

Removed the `@AfterAll spark.stop()` calls from the three standalone tests and converted them to proper `@SpringBootUnitTest` tests that receive the shared `SparkSession` via Spring injection, consistent with all other Spark-dependent tests in the server module. The Spring-managed session is never stopped by individual tests.

### LINKED ISSUE #2617: Fix: SearchProviderAuthTest flaky due to leaked SecurityContext mock in BulkSubmitProviderTest
`SearchProviderAuthTest.searchSucceedsWithCorrectResourceAuthority` fails intermittently when run in parallel with `BulkSubmitProviderTest`.

## Root cause

`BulkSubmitProviderTest.setUp()` installs a Mockito mock as the active Spring `SecurityContext` via `SecurityContextHolder.setContext(securityContext)` but never calls `SecurityContextHolder.clearContext()` after each test. Because JUnit 5 parallel execution schedules tests on a shared `ForkJoinPool`, a subsequent test that happens to run on the same worker thread will inherit the leaked mock context.

When `SearchProviderAuthTest` runs on such a thread:
1. `SearchProviderAuthTest.setSecurityContext()` calls `SecurityContextHolder.getContext().setAuthentication(auth)` — but `getContext()` returns the Mockito mock, and `setAuthentication()` on a mock is a no-op.
2. `SecurityContextHolder.getContext().getAuthentication()` then returns the original mock `Authentication` object, which is **not** an `AbstractAuthenticationToken`.
3. `checkHasAuthority()` detects this and throws `AccessDeniedError: Token not present`.

## Error extract

```
[ERROR] Errors:
[ERROR]   SearchProviderAuthTest.searchSucceedsWithCorrectResourceAuthority:102 » AccessDenied Token not present

au.csiro.pathling.errors.AccessDeniedError: Token not present
    at au.csiro.pathling.security.PathlingAuthorities.checkHasAuthority(PathlingAuthorities.java:91)
    at au.csiro.pathling.search.SearchProvider.search(SearchProvider.java:117)
    at au.csiro.pathling.search.SearchProviderAuthTest.searchSucceedsWithCorrectResourceAuthority(SearchProviderAuthTest.java:102)
    at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
    at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
    ...
```

## Fix

Add `@AfterEach void tearDown() { SecurityContextHolder.clearContext(); }` to `BulkSubmitProviderTest` so the mock context is removed after each test method, preventing leakage to other tests running on the same thread.

### LINKED ISSUE #2620: fix: $import-pnp rejects its own staged files on local filesystem
## Summary

`$import-pnp` against a local-filesystem staging directory fails after the bulk export has completed, with a 4xx response carrying:

```
Ping and pull import failed: URL: 'file:///<staging-dir>/pathling-pnp-import-<jobId>/export-output/<Resource>.0000.ndjson' is not an allowed source.
```

The downloaded files are inside the staging directory the executor created and trusts, but the allowlist check it generates for that directory does not match the URLs of those files.

## Cause

`ImportPnpExecutor.execute` constructs the staging allowlist prefix with:

```java
final List<String> pnpAllowableSources = List.of(tempDir.toUri().toString() + "/");
```

where `tempDir` is a Hadoop `org.apache.hadoop.fs.Path` built via `new Path(baseDir, tempDirName)`. For a local filesystem this drops the empty authority and yields `file:/<path>/` (one slash). The downloaded file URLs are reported by `fs.listFiles + fs.makeQualified`, which preserves the empty authority and produces `file:///<path>/<file>` (three slashes). `UrlAllowlist` falls back to case-sensitive `String.startsWith` for non-HTTP schemes, so the two forms no longer match and `ImportExecutor.checkCustomAllowableSources` rejects every downloaded file with an `AccessDeniedError`.

The behaviour was introduced by `b6ec9b5c74 refactor: Stage \$import-pnp downloads via Hadoop FileSystem`, which switched the prefix from `"file://" + tempDir.toAbsolutePath()` (which produced the three-slash form) to `tempDir.toUri().toString()`. Non-local staging schemes (s3a, hdfs) embed the bucket/namenode as authority and are unaffected — this only breaks plain local staging.

## Reproduction

Configure `pathling.import.pnp.downloadLocation` to a local-filesystem path and run a successful `\$import-pnp` against any bulk export endpoint. The kick-off and download phases complete (visible in WireMock or real-server logs), then the operation fails when handing the files to `ImportExecutor`. Verified by `ImportPnpOperationIT.testMissingInputSourceSucceeds` and `testValidRequestReturns202WithContentLocation`.

## Fix

Build the prefix via `fs.getFileStatus(tempDir).getPath().toUri().toString()` so both sides use the same canonical URI form, preserving the empty authority on `file://` paths. Already shipped in commit `906ef4933e` on the branch attached to this issue.

### LINKED ISSUE #2621: test: $import-pnp and $bulk-submit integration tests have stale-state and config gaps
## Summary

Three independent integration-test problems landed together and were each producing 4xx responses, masking each other and (in one case) hiding the production bug tracked in #2620:

1. `ImportPnpOperationIT` leaks Delta state across tests, so any test that imports after a previous import gets `DELTA_PATH_EXISTS` instead of the success or error the test is exercising.
2. `ImportPnpOperationIT.testPoisonedManifestTypeFailsJobAndBlocksExfiltration` sends an unauthenticated request even though the class enables auth and configures PNP credentials. The class-level interlock returns 401 before the poisoning path executes.
3. `BulkSubmitOperationIT` and `BulkSubmitOAuthIT` configure `pathling.bulk-submit.allowable-sources[0]=http://localhost` via `@TestPropertySource`, which the URI-aware `UrlAllowlist` resolves to effective port 80 and no longer matches the dynamic WireMock port.

### 1. Delta cache leak between PNP tests

The class uses a static `@TempDir warehouseDir` and an `@AfterEach` that calls `FileUtils.cleanDirectory(warehouseDir.toFile())`. The disk is cleaned, but Spark's catalog cache and Delta's global `DeltaLog` cache still hold metadata for tables that were just deleted. The next test rebuilds the warehouse from fixtures via `TestDataSetup.copyTestDataToTempDir`, but `DeltaTable.isDeltaTable` consults the stale log and returns `false`. `DeltaSink` therefore falls through to `writeDataset(..., SaveMode.ERROR_IF_EXISTS)`, which collides with the freshly-copied directory and fails:

```
[DELTA_PATH_EXISTS] Cannot write to already existent path file:/<warehouse>/delta/<Resource>.parquet without setting OVERWRITE = 'true'.
```

Reproduce by running `ImportPnpOperationIT` end-to-end after the #2620 fix is in place: an early test (e.g. `testValidRequestReturns202WithContentLocation`) succeeds, and a later test importing the same resource type (`testMissingInputSourceSucceeds`) hits `DELTA_PATH_EXISTS`.

### 2. Missing auth header

`testPoisonedManifestTypeFailsJobAndBlocksExfiltration` omits `Authorization: Bearer <mock-token>` on the POST and on both follow-up GETs (`\$job` poll and `/jobs/<id>/<file>` exfiltration probe). The class is annotated `pathling.auth.enabled=true` and configures PNP credentials, so `ImportPnpOperationValidator.validateAuthConfiguration` returns 401 before the executor ever sees the poisoned manifest. The pre-existing 401 was masked because the test was already failing earlier in the request path before the #2620 fix.

### 3. Bare `http://localhost` allowable source

`BulkSubmitOperationIT` and `BulkSubmitOAuthIT` declare:

```java
@TestPropertySource(properties = {
  ...
  \"pathling.bulk-submit.allowable-sources[0]=http://localhost\",
  ...
})
```

`UrlAllowlist.matches` parses `http://localhost` as a URI and treats it as effective port 80. The tests target WireMock on a random port, so manifest URLs like `http://localhost:<wireMockPort>/manifest.json` no longer match and the operation returns 400 \"manifest URL does not match any allowed source prefixes\". The fix is to register the property in `@DynamicPropertySource` so it can include `wireMockServer.port()` at runtime — the same pattern the same classes already use for `warehouseUrl` and `staging-directory`.

## Fixes

All three are fixed on the PR attached to this issue:

- `ca1ce49c07` — Clear `pathlingContext.getSpark().catalog()` and `DeltaLog.clearCache()` in `@AfterEach` before deleting files.
- `780db513ed` — Add `Authorization: Bearer <AUTH_TOKEN>` on the three requests inside the poisoned-manifest test.
- `aa4fdc6862` — Move `allowable-sources[0]` from `@TestPropertySource` into `@DynamicPropertySource`, using `wireMockServer.port()`.

piotrszul:
## Summary

This PR groups four fixes for integration- and unit-test failures that stem from shared-state leaks and a real production bug exposed by recent refactors.

### Unit-test isolation

- Standalone Spark tests (`SqlQueryResultStreamerTest`, `ViewRegistrationServiceTest`, `LibraryReferenceResolverTest`) called `spark.stop()` in `@AfterAll`, destroying the JVM-wide `SparkContext` and causing subsequent tests to fail with `No active or default Spark session found`. Fixed by converting them to `@SpringBootUnitTest` so they share the managed Spring `SparkSession` bean and never call `stop()`.
- `BulkSubmitProviderTest` installed a Mockito mock as the active Spring `SecurityContext` in `@BeforeEach` but had no `@AfterEach` to clear it. Under JUnit 5 parallel execution the mock leaked onto adjacent `ForkJoinPool` threads, causing `SearchProviderAuthTest` to inherit the mock context and fail with `AccessDeniedError: Token not present`. Fixed by adding `@AfterEach void tearDown() { SecurityContextHolder.clearContext(); }`.

### `$import-pnp` staging allowlist (#2620)

`ImportPnpExecutor` built its staging-directory allowlist prefix with `tempDir.toUri().toString()`, which drops the empty authority on `file://` paths (yielding `file:/path`) while downloaded files come back from `fs.makeQualified` as `file:///path`. `UrlAllowlist`'s string-prefix match then rejected every downloaded file. Fixed by going through `fs.getFileStatus(tempDir).getPath()` so both sides use the same canonical URI form.

### `$import-pnp` / `$bulk-submit` integration-test fixes (#2621)

- `ImportPnpOperationIT` cleans `warehouseDir` in `@AfterEach` but leaves stale entries in Spark's catalog cache and Delta's global `DeltaLog` cache. Subsequent tests rebuild the warehouse from fixtures, but Delta's `isDeltaTable` returns false against the stale log, sending the import down `ERROR_IF_EXISTS` and failing with `DELTA_PATH_EXISTS`. Fixed by clearing both caches before deleting files.
- `testPoisonedManifestTypeFailsJobAndBlocksExfiltration` sent its requests without the `Authorization: Bearer <token>` header; the class-level auth interlock rejected them with 401 before the poisoning path executed.
- `BulkSubmitOperationIT` and `BulkSubmitOAuthIT` had `pathling.bulk-submit.allowable-sources[0]=http://localhost` in `@TestPropertySource`. The URI-aware `UrlAllowlist` resolves that prefix to effective port 80 and no longer matches the dynamic WireMock port. Moved the property into `@DynamicPropertySource` so it includes `wireMockServer.port()` at runtime.

Closes #2615.
Closes #2617.
Closes #2620.
Closes #2621.

## Test plan

- [x] Reproduced Spark failures by observing `ViewDefinitionSearchTest` and `ViewDefinitionCreateTest` errors in full suite run
- [x] Reproduced `SearchProviderAuthTest` failure with `forkCount=1`
- [x] Reproduced `ImportPnpOperationIT` failures in isolation (URL-allowlist bug) and in suite (Delta cache leak)
- [x] Full suite passes with both fixes applied (`forkCount=2`): 1481 tests, 0 failures, 0 errors
- [x] `mvn verify -Dit.test='ImportPnpOperationIT,BulkSubmitOperationIT,BulkSubmitOAuthIT'` passes: 15 tests, 0 failures, 0 errors

### DIFF
--- a/server/src/main/java/au/csiro/pathling/operations/bulkimport/ImportPnpExecutor.java
+++ b/server/src/main/java/au/csiro/pathling/operations/bulkimport/ImportPnpExecutor.java
@@ -158,9 +158,12 @@ public class ImportPnpExecutor {
 
       // Execute the import using the existing ImportExecutor with custom allowable sources.
       // This bypasses the configured allowableSources validation for the staging directory,
-      // which the server downloaded and trusts. The qualified URI ensures the prefix matches
-      // whatever scheme the staging file system uses.
-      final List<String> pnpAllowableSources = List.of(tempDir.toUri().toString() + "/");
+      // which the server downloaded and trusts. Go via fs.getFileStatus() to obtain a URI in
+      // the same canonical form (with empty authority preserved on file://) that fs.listFiles
+      // produces for the downloaded files, so the UrlAllowlist string-prefix match holds:
+      // tempDir.toUri() alone yields file:/path while listed files come back as file:///path.
+      final List<String> pnpAllowableSources =
+          List.of(fs.getFileStatus(tempDir).getPath().toUri().toString() + "/");
       final ImportResponse response =
           importExecutor.execute(importRequest, jobId, pnpAllowableSources);
 
--- a/server/src/test/java/au/csiro/pathling/operations/bulkimport/ImportPnpOperationIT.java
+++ b/server/src/test/java/au/csiro/pathling/operations/bulkimport/ImportPnpOperationIT.java
@@ -701,6 +701,7 @@ class ImportPnpOperationIT {
             .header("Content-Type", "application/fhir+json")
             .header("Accept", "application/fhir+json")
             .header("Prefer", "respond-async")
+            .header("Authorization", "Bearer " + AUTH_TOKEN)
             .bodyValue(requestBody)
             .exchange()
             .expectStatus()
@@ -736,6 +737,7 @@ class ImportPnpOperationIT {
                     .get()
                     .uri(contentLocation)
                     .header("Accept", "application/fhir+json")
+                    .header("Authorization", "Bearer " + AUTH_TOKEN)
                     .exchange()
                     .expectStatus()
                     .is4xxClientError()
@@ -752,6 +754,7 @@ class ImportPnpOperationIT {
     webTestClient
         .get()
         .uri("http://localhost:" + port + "/jobs/" + jobId + "/escaped.0000.ndjson")
+        .header("Authorization", "Bearer " + AUTH_TOKEN)
         .exchange()
         .expectStatus()
         .isNotFound();
--- a/server/src/test/java/au/csiro/pathling/operations/bulkimport/ImportPnpOperationIT.java
+++ b/server/src/test/java/au/csiro/pathling/operations/bulkimport/ImportPnpOperationIT.java
@@ -173,6 +173,11 @@ class ImportPnpOperationIT {
 
   @AfterEach
   void cleanup() throws IOException {
+    // Clear cached Delta table state before deleting files. Otherwise the next test sees a stale
+    // DeltaLog in memory that no longer matches the on-disk warehouse rebuilt from test fixtures,
+    // and Delta refuses the import with DELTA_PATH_EXISTS.
+    pathlingContext.getSpark().catalog().clearCache();
+    org.apache.spark.sql.delta.DeltaLog.clearCache();
     FileUtils.cleanDirectory(warehouseDir.toFile());
   }
 
--- a/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitOAuthIT.java
+++ b/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitOAuthIT.java
@@ -71,7 +71,6 @@ import org.springframework.test.web.reactive.server.WebTestClient;
 @TestPropertySource(
     properties = {
       "pathling.async.enabled=true",
-      "pathling.bulk-submit.allowable-sources[0]=http://localhost",
       // Configure submitter with OAuth credentials for symmetric (client_secret) auth.
       "pathling.bulk-submit.allowed-submitters[0].system=http://example.org/submitters",
       "pathling.bulk-submit.allowed-submitters[0].value=oauth-submitter",
@@ -118,6 +117,11 @@ class BulkSubmitOAuthIT {
     TestDataSetup.copyTestDataToTempDir(warehouseDir, "Condition");
     registry.add("pathling.storage.warehouseUrl", () -> "file://" + warehouseDir.toAbsolutePath());
     registry.add("pathling.bulk-submit.staging-directory", stagingDir::toString);
+    // The allowable source must include the WireMock port. Bare "http://localhost" (port 80)
+    // no longer matches "http://localhost:{port}" under the URI-aware UrlAllowlist matching.
+    registry.add(
+        "pathling.bulk-submit.allowable-sources[0]",
+        () -> "http://localhost:" + wireMockServer.port());
   }
 
   @BeforeEach
--- a/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitOperationIT.java
+++ b/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitOperationIT.java
@@ -67,7 +67,6 @@ import org.springframework.test.web.reactive.server.WebTestClient;
 @TestPropertySource(
     properties = {
       "pathling.async.enabled=true",
-      "pathling.bulk-submit.allowable-sources[0]=http://localhost",
       "pathling.bulk-submit.allowed-submitters[0].system=http://example.org/submitters",
       "pathling.bulk-submit.allowed-submitters[0].value=test-submitter"
     })
@@ -107,6 +106,11 @@ class BulkSubmitOperationIT {
     TestDataSetup.copyTestDataToTempDir(warehouseDir, "Condition");
     registry.add("pathling.storage.warehouseUrl", () -> "file://" + warehouseDir.toAbsolutePath());
     registry.add("pathling.bulk-submit.staging-directory", stagingDir::toString);
+    // The allowable source must include the WireMock port. Bare "http://localhost" (port 80)
+    // no longer matches "http://localhost:{port}" under the URI-aware UrlAllowlist matching.
+    registry.add(
+        "pathling.bulk-submit.allowable-sources[0]",
+        () -> "http://localhost:" + wireMockServer.port());
   }
 
   @BeforeEach
--- a/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitProviderTest.java
+++ b/server/src/test/java/au/csiro/pathling/operations/bulksubmit/BulkSubmitProviderTest.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Optional;
 import org.hl7.fhir.r4.model.Parameters;
 import org.hl7.fhir.r4.model.StringType;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.security.core.Authentication;
@@ -79,6 +80,11 @@ class BulkSubmitProviderTest {
     final Authentication authentication = mock(Authentication.class);
     when(securityContext.getAuthentication()).thenReturn(authentication);
     SecurityContextHolder.setContext(securityContext);
+  }
+
+  @AfterEach
+  void tearDown() {
+    SecurityContextHolder.clearContext();
   }
 
   // ========================================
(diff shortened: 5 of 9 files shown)


johngrimes (approved):
Looks good to me!

PR #2628, opened 2026-06-12
Add command line interface

johngrimes:
This adds a `pathling` console script to the Python package, surfacing the library's functionality through a flat, verb-based command tree. It covers data conversion, SQL on FHIR views, FHIRPath evaluation, bulk export, and terminology operations over code datasets. The CLI is built with Click and Rich, resolves configuration from flags then an optional TOML file, and is installable and runnable via `uv tool` / `uvx` so its version always matches the library.

The package's public names are now exposed lazily, so importing the CLI doesn't pull in PySpark and `--help`/`--version` stay fast. Spark and JVM logging is suppressed by default (and re-enabled with `--verbose`), errors are rendered as concise messages without Java stack traces, data goes to stdout while progress and errors go to stderr, and the Spark worker interpreter is pinned to the driver to avoid version mismatches.

Commands: `convert`, `view`, `fhirpath`, `export`, and the terminology operations `member-of`, `translate`, `subsumes`, `subsumed-by`, `display`, `property-of`, and `designation`.

Tested with unit tests for the parsing, config, rendering, and error-mapping logic, per-command integration tests through Click's runner against the existing Spark and mock-server fixtures, and a subprocess smoke test of the installed entry point. Documentation is added under `site/docs/libraries` and the Python README.

### DIFF
--- a/lib/R/pom.xml
+++ b/lib/R/pom.xml
@@ -227,6 +227,15 @@
                 <argument>install.packages("remotes", type="source")</argument>
                 <argument>-e</argument>
                 <argument>devtools::install_dev_deps(type="source")</argument>
+                <!-- Pin dbplyr to the last release before 2.6.0. dbplyr 2.6.0 is
+                     incompatible with sparklyr 1.9.4: its query-fields probe
+                     emits standard SQL (double-quoted identifiers) that the
+                     Spark parser rejects, breaking all tbl_spark operations.
+                     Run after install_dev_deps so the pin holds regardless of
+                     what the package cache restored. Remove once a sparklyr
+                     release supports dbplyr 2.6.0. -->
+                <argument>-e</argument>
+                <argument>remotes::install_version("dbplyr", version = "2.5.2", type = "source", upgrade = "never")</argument>
                 <argument>-e</argument>
                 <argument>installed.packages()[,c("Package","Version")]</argument>
                 <argument>-e</argument>
--- a/lib/python/README.md
+++ b/lib/python/README.md
@@ -22,8 +22,9 @@ pip install pathling
 
 The package ships a `pathling` console script that surfaces the library's
 functionality - data conversion, SQL on FHIR views, FHIRPath evaluation, bulk
-export, and terminology operations - through scriptable commands. It requires a
-supported Java runtime, as does the library itself.
+export, terminology operations, Python scripting (`run`), and an interactive
+console (`console`) - through scriptable commands. It requires a supported
+Java runtime, as does the library itself.
 
 The easiest way to run it is with [uv](https://docs.astral.sh/uv/):
 
@@ -42,6 +43,16 @@ view:
 ```bash
 pathling convert data/ --to parquet -o warehouse/
 pathling view data/ --view patients.json --format csv
+```
+
+The `run` and `console` commands execute Python code (a script, stdin, or
+`-c CODE`) or open an interactive IPython session, with `spark` and `pathling`
+variables already configured and in scope:
+
+```bash
+pathling run my_script.py
+pathling run -c "print(spark.version)"
+pathling console
 ```
 
 See the [command line interface documentation](https://pathling.csiro.au/docs/libraries/cli)
--- a/lib/python/README.md
+++ b/lib/python/README.md
@@ -1,5 +1,4 @@
-Python API for Pathling
-=======================
+# Python API for Pathling
 
 This is the Python API for [Pathling](https://pathling.csiro.au). It provides a
 set of tools that aid the use of FHIR terminology services and FHIR data within
@@ -16,8 +15,37 @@ Prerequisites:
 To install, run this command:
 
 ```bash
-pip install pathling  
+pip install pathling
 ```
+
+## Command line interface
+
+The package ships a `pathling` console script that surfaces the library's
+functionality - data conversion, SQL on FHIR views, FHIRPath evaluation, bulk
+export, and terminology operations - through scriptable commands. It requires a
+supported Java runtime, as does the library itself.
+
+The easiest way to run it is with [uv](https://docs.astral.sh/uv/):
+
+```bash
+# Run without installing.
+uvx pathling --version
+
+# Or install it as a tool.
+uv tool install pathling
+pathling --help
+```
+
+For example, to convert a directory of NDJSON to Parquet, or run a SQL on FHIR
+view:
+
+```bash
+pathling convert data/ --to parquet -o warehouse/
+pathling view data/ --view patients.json --format csv
+```
+
+See the [command line interface documentation](https://pathling.csiro.au/docs/libraries/cli)
+for the full list of commands and options.
 
 ## Encoders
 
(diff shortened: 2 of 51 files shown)


piotrszul (commented):
Reviewed at `7a105921fd`. Consolidates three independent passes (two manual + one fresh reviewer with no shared context) plus empirical verification where noted (✅). The CLI is well-structured — clean module separation, fail-fast validation before Spark cold-start, careful stdout/stderr discipline, and genuinely fast `--help`/`--version` via the lazy import shim. The items below are prioritized; most are pre-merge fixes, with two enhancement requests at the end.

## 🔴 Should fix before merge

**1. Broad `except` mislabels every export failure as an auth failure** — `cli/export.py` (`try/except` around `pc.read.bulk(...)`)
When auth is configured, the catch-all re-raises *any* failure (server down, timeout, HTTP 500 during kickoff/polling/download) as "Authentication failed … check the client ID, credential, and token endpoint." Misdiagnoses network/server errors as bad credentials. → Narrow the catch to auth/401, or include the unwrapped root cause without asserting it's an auth problem.

**2. Mistyped `--config <missing>` silently falls back to the default config's `[bulk-auth]`** — `cli/export.py` + `cli/config.py`
`resolve_config` nulls `config_path` when the path doesn't exist; export then does `config.config_path or default_config_path()`, so `--config /typo.toml` makes *export* authenticate with `~/.config/pathling/config.toml`'s credentials while the rest of the CLI read nothing from the explicit file. Silent, security-relevant. → Error if an explicit `--config` path doesn't exist; reuse the already-resolved config.

**3. Incomplete bulk-auth runs the export unauthenticated** ✅ — `cli/config.py` `resolve_bulk_auth`
Supplying `--client-secret`/`--private-key-jwk` + `--token-endpoint` but omitting `--client-id` returns `None`, so export runs with **no auth** and fails later with an opaque 401 instead of "you forgot `--client-id`." → If any bulk-auth flag/secret is given but `client-id` is missing, raise a usage error.

**4. Lazy exports break `pathling.<submodule>.X` attribute access** ✅ — `pathling/__init__.py`
The PEP 562 shim dropped the eager `from .functions import …` / `from .udfs import …`, so `import pathling; pathling.udfs.member_of` now raises `AttributeError` (it "heals" only after a lazy name from that submodule, or `import pathling.udfs`, runs). The documented top-level API (`from pathling import member_of`) is unaffected, so blast radius depends on downstream submodule-attribute usage — but it's a silent, order-dependent regression of a previously-valid pattern. → In `__getattr__`, fall back to importing a known submodule and cache it; add those names to `__dir__`.

## 🟠 Worth fixing

**5. Test suite isn't portable to Python ≥ 3.10** ✅ — `tests/cli/conftest.py`
`CliRunner(mix_stderr=False)`: the kwarg was removed in Click 8.2.0 (which also dropped Python 3.9). The lockfile resolves Click 8.1.8 for `<3.10` and 8.3.1 for `≥3.10`. CI stays green because the build venv is pinned to 3.9 (`pom.xml -p 3.9`, `.python-version`), but the suite errors at fixture setup on any modern interpreter. Runtime is unaffected (test-only). → Feature-detect: pass `mix_stderr=False` only if the parameter exists in `CliRunner.__init__`.

**6. `friendly_message`'s connection branch is dead code** — `cli/main.py` / `cli/errors.py`
The central handler never passes `server_url`, so "Could not reach the server at <url>" never fires for `convert`/`view`/`fhirpath`; they fall through to the generic "--verbose" message. → Thread the configured `tx_server` in, or remove the branch.

**7. Post-write summaries re-scan output with per-type `count()`** — `cli/export.py`, `cli/convert.py` `_print_summary`
After the read+write pass, the summary calls `.read(rt).count()` per resource type — N extra full Spark jobs just to print row counts. → Derive counts from the write, or make the summary opt-in.

**8. Bundles are parsed twice** — `cli/io.py`
`discover_bundle_resource_types` `json.load`s every bundle on the driver (single-threaded) to enumerate types, then `pc.read.bundles` reparses them in Spark; `_looks_like_bundle` also full-parses the first JSON file (the XML branch reads only a prefix), and discovery runs outside the progress spinner. → Read a prefix in `_looks_like_bundle`; stream discovery / run it under the spinner / add `--type` to skip it; longer-term let the library reader discover types.

**9. IPython/rich are now hard runtime deps of the whole library** — `pyproject.toml`
`ipython>=8.18.1` (CLI-only, used by `console`) is pulled into every `pip install pathling`, even for API-only users. → Move CLI deps to an optional extra (`pathling[cli]`) and import-guard `console`.

**10. Quiet log4j2 temp file is never removed** — `cli/session.py` `_write_quiet_log4j2`
`NamedTemporaryFile(delete=False)` with no cleanup; every non-verbose run leaks one `pathling-cli-log4j2-*.properties` into the temp dir. → `atexit`-unlink, or (since the content is constant) write once to a cache path / ship as package data.

**11. File output fully materialises on the driver; "stream" wording is inaccurate** — `cli/render.py`
stdout csv/json/ndjson use `df.collect()` and `-o` uses `df.toPandas()` — both buffer the whole result in driver memory, but the docstring and docs say csv/json/ndjson "stream the full result." (The `--limit`-applies-to-table-only behaviour is **by design and correctly documented** — not a bug.) → Fix the "stream" wording; optionally use `toLocalIterator()` for incremental stdout. *(See enhancement E-2 below for the scalable-output side of this.)*

## 🔵 Minor / nits
- Terminology `_validate_columns` runs after Spark cold-start (a typo'd `--code-column` only errors post-startup) — `cli/terminology.py`.
- `_categorise` keys "fhirpath" off the broad substring `"parse error"` — `cli/errors.py`.
- Partial terminology auth is silently disabled (no warning) unless both `client_id` + `token_endpoint` are set — `cli/config.py`.
- Hand-built Coding struct duplicates the library's Coding schema (drift risk) — `cli/terminology.py`.
- Unrelated README/Databricks-doc markdown reformatting inflates the diff.

## 🟢 Enhancement requests (not blockers)

**E-1. Make the Spark session configurable via the config file (esp. driver memory).** There's currently no way to set `spark.driver.memory` or other Spark options. Suggest a `[spark]` table in the existing TOML config, threaded into `_build_spark_session` alongside `tx-server`/`[*-auth]`, e.g.:
```toml
[spark]
"spark.driver.memory" = "8g"
```
with optional `--conf key=value` overrides. Note: `spark.driver.memory` must be applied before the driver JVM launches, so verify the mechanism takes effect in local mode.

**E-2. `view`: support both single-file and partitioned Spark output.** `view -o` is currently always single-file (`toPandas()`); offer a `--single-file/--partitioned` (or `--coalesce N`) choice that routes partitioned writes through the Spark writer for the formats that support it (Parquet/Delta/CSV/JSON), as `convert` already does. This is the scalable counterpart to #11.

## Assessment
**Merge with fixes.** Strongest items are the error/auth-correctness bugs (#1–#3) and the API regression (#4); #5 unblocks contributors on modern Python. The rest are quality/perf improvements and two enhancement requests.

---
<sub>🤖 Review assembled with [Claude Code](https://claude.com/claude-code) assistance, with findings verified against the code where marked ✅.</sub>

piotrszul (approved):
Looks good to me.

PR #2630, opened 2026-06-18
Make FHIRPath evaluation safe under Spark ANSI mode

### LINKED ISSUE #2629: Support for default ANSI mode
Writing transform SQL that stays portable across multiple database engines effectively requires assuming lenient cast behaviour, where an invalid or empty value casts to NULL rather than raising. Lenient casting was the common default across engines for a long time, and relying on it avoids baking each vendor's non-standard "safe cast" spelling (`TRY_CAST`, `SAFE_CAST`, and so on) into the SQL per engine. That assumption sits at odds with Pathling's default Spark settings, and presumably with what Pathling works against and is tested against: Spark 4 enables ANSI mode (`spark.sql.ansi.enabled=true`) by default, under which those casts raise. Running portable SQL on the same session as Pathling therefore currently means forcing `spark.sql.ansi.enabled=false`.

The open question is the knock-on effect on Pathling itself. Because the setting is session-wide, it also relaxes the casts Pathling performs when it evaluates ViewDefinitions and maps FHIRPath results onto typed columns. A few questions:

- Is Pathling 9.7.x tested and supported on Spark 4 with `spark.sql.ansi.enabled` left at its default (`true`)?
- Does view evaluation rely anywhere on lenient cast semantics (for example, casting an empty or non-conforming primitive to its declared column type), such that ANSI mode could surface errors that don't appear today?
- If ANSI mode is expected to be fine, are there particular column types or FHIRPath constructs to be careful with?

The main concern is whether disabling ANSI mode for the portable SQL might be masking a Pathling incompatibility that a consumer on a stock Spark 4 session would otherwise hit. Happy to provide a minimal reproduction if that would help.

Environment: Pathling 9.7.1, PySpark 4.0.3, `local[*]`, Hive-enabled session.

johngrimes:
Spark 4 enables `spark.sql.ansi.enabled` by default, which makes invalid or out-of-range casts and numeric overflow raise errors rather than returning empty. In practice that meant a single non-conforming value in real-world FHIR data could abort an entire query. It also left open the concern raised in the issue: consumers who disable ANSI mode to run portable transform SQL on the same session might be unknowingly masking a Pathling incompatibility.

This makes the core `fhirpath` and `encoders` evaluation produce identical, correct results regardless of the setting. Failure-prone decimal and Quantity casts now route through a single safe-cast helper built on `try_cast`, the `+`/`-`/`*` operators use their `try_*` counterparts so overflow yields empty, and `count`/`isEmpty`/`toBoolean` no longer rely on `size(null)` (which Spark resolves differently depending on the flag). On failure the result is empty/NULL under both settings, matching FHIRPath and SQL on FHIR semantics, and no production code reads or sets the flag.

A new parameterised test suite evaluates the divergent cases under both settings and asserts identical results, and the test harness now asserts it starts under ANSI-on so the supported default cannot regress unnoticed. The full `fhirpath` and `encoders` suites pass with ANSI mode both on and off. The Spark configuration documentation now records that either setting is supported.

Closes #2629.

### DIFF
--- a/encoders/src/test/java/au/csiro/pathling/encoders/AnsiTestSupport.java
+++ b/encoders/src/test/java/au/csiro/pathling/encoders/AnsiTestSupport.java
@@ -37,16 +37,32 @@ public final class AnsiTestSupport {
   private AnsiTestSupport() {}
 
   /**
-   * Disables ANSI mode on the session when {@code pathling.test.ansiEnabled} is set to {@code
-   * false}; otherwise leaves the session at the Spark 4 default (ANSI on).
+   * Reconciles the supported ANSI default with the optional ANSI-off verification run.
+   *
+   * <p>When the {@code pathling.test.ansiEnabled} system property is set to {@code false}, ANSI
+   * mode is disabled on the session so the suite can be run once under the lenient setting
+   * (FR-011). Otherwise the property is treated as unset: the flag is left at the Spark 4 default
+   * and the resolved value is asserted to be {@code true}, failing loudly so the supported default
+   * cannot regress unnoticed (FR-009).
    *
    * @param spark the session to configure
    * @return the same session, for fluent use directly on a {@code getOrCreate()} call
+   * @throws IllegalStateException if ANSI mode is expected to be enabled but is not
    */
   @Nonnull
   public static SparkSession configureAnsiMode(@Nonnull final SparkSession spark) {
     if ("false".equalsIgnoreCase(System.getProperty("pathling.test.ansiEnabled"))) {
+      // Explicit ANSI-off verification run: relax the session-wide flag.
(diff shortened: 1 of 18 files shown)


piotrszul (approved):
Looks good to me.
I have committed three minor changes.

Clarify that the size(null) guards in count/isEmpty/toBoolean make the
result independent of both spark.sql.legacy.sizeOfNull and
spark.sql.ansi.enabled, rather than implying ANSI-off alone diverges.

Mirror the fail-loud ANSI assertion from the fhirpath harness into the
encoders test support so the encoders suite cannot silently run under
the wrong mode if a Spark default ever changes.

Remove the unused sum() representation and its test: it was the last
remaining raw arithmetic operator and had no production callers.

PR #2637, opened 2026-06-25
feat: implemented `startsWith`, `endsWith`, and `contains` string functions

chgl:
The Java code was Claude-generated, the Python test hand-crafted.


fhirpath/src/test/java/au/csiro/pathling/fhirpath/dsl/StringFunctionsDslTest.java
```diff
@@ -78,4 +79,99 @@ public Stream<DynamicTest> testJoin() {
...
+            "singleString.startsWith('Hello, world!')",
+            "startsWith() returns true when prefix equals the whole string")
+        .group("startsWith() function with empty values")
+        .testEmpty("emptyString.startsWith('Hello')", "startsWith() on empty input returns empty")
+        .testEmpty(
+            "singleString.startsWith({})", "startsWith() with empty prefix argument returns empty")
+        .group("startsWith() function error cases")
+        .testError(
+            "stringArray.startsWith('o')",
+            "startsWith() errors when input collection is not singular")
```
johngrimes, line 112: The error cases cover a multi-item input collection (`stringArray.startsWith('o')`), but not a multi-item argument, e.g. `singleString.startsWith(stringArray)`.

The spec requires the argument to be a singleton string, and the implementation handles this correctly via `asSingular()` on the argument - it would just be good to have a test pinning that behaviour down for each of the three functions.

fhirpath/src/main/java/au/csiro/pathling/fhirpath/function/provider/StringFunctions.java
```diff
@@ -67,4 +70,76 @@ public static StringCollection join(
                     ? separator.asSingular().getColumn()
                     : DefaultRepresentation.literal(JOIN_DEFAULT_SEPARATOR)));
   }
+
+  /**
+   * The startsWith function returns true when the input string starts with the given prefix.
+   *
+   * <p>If prefix is the empty string ({@code ''}), the result is {@code true}.
+   *
+   * <p>If the input collection or prefix is empty, the result is empty.
```
johngrimes: Suggest:

> If the input collection or the prefix argument is an empty collection ({ }), the result is empty.

johngrimes (changes requested):
Thanks @chgl, looks good to me!

Just a couple of minor issues that I noticed.

Thanks very much for your contribution!

johngrimes (approved):

johngrimes:
Sorry @chgl, I meant to release this but I must have forgot to merge it.

I will make sure that this goes out in the next release.

johngrimes:
Thanks @chgl - this has been merged into the 9.10.0 release branch, and a pre-release Python package containing these functions is now available on PyPI:

```
pip install pathling==9.10.0.dev0
```

Published from the Python pre-release workflow on `release/9.10.0`: https://github.com/aehrc/pathling/actions/runs/34454273457

chgl:
Great, thank you :)

PR #2665, opened 2026-07-23
Review changes for 9.9.0

piotrszul:
## Summary
Review branch for the 9.9.0 release cycle, aggregating recent changes for review against `release/9.9.0` ahead of the release PR into `main`, #2656.

Small, cosmetic changes found during review are included directly as commits on this branch. Other review feedback is posted as comments below, to be reviewed and triaged separately.

### DIFF
--- a/lib/python/pathling/cli/console.py
+++ b/lib/python/pathling/cli/console.py
@@ -40,20 +40,24 @@ def build_banner() -> str:
     """Builds the banner shown before the console's first prompt.
 
     The banner identifies the Pathling and Python versions, lists the
-    variables in scope, notes that the Pathling public functions are
-    pre-imported (with the terminology display available as ``tx_display``),
-    and explains how to exit.
+    variables in scope, notes that the Pathling public functions (e.g.
+    ``member_of``, ``translate``, ``to_coding``) are pre-imported, points to
+    the Python API reference for the full list, notes that the terminology
+    display is available as ``tx_display`` rather than ``display``, and
+    explains how to exit.
 
     :return: the banner text.
     """
     return (
         f"Pathling console (version {__version__}, "
         f"Python {platform.python_version()})\n"
-        "Variables in scope: spark (SparkSession), pathling (PathlingContext)\n"
-        "Pathling public functions are pre-imported (member_of, translate, "
-        "to_coding, ...); see https://pathling.csiro.au/docs/python/pathling.html\n"
-        "The terminology display function is available as tx_display "
-        "(display is IPython's built-in).\n"
+        "Variables in scope: `spark` (SparkSession), `pathling` (PathlingContext)\n"
+        "Pathling public functions are pre-imported (`member_of`, `translate`, "
+        "`to_coding`, ...).\n"
+        "See https://pathling.csiro.au/docs/python/pathling.html#module-pathling "
+        "for the full list.\n"
+        "The terminology display function is available as `tx_display` "
+        "(`display` is IPython's built-in).\n"
         "Type exit or press Ctrl-D to leave.\n"
     )
 
--- a/lib/python/pathling/cli/render.py
+++ b/lib/python/pathling/cli/render.py
@@ -179,7 +179,8 @@ def output_options(func):
             "delimiter",
             default=",",
             callback=_delimiter_callback,
-            help="Field separator for CSV input and output (default: ',').",
+            help="Field separator for CSV input and output, including .tsv "
+            "(default: ','; not inferred from the file extension).",
         ),
         click.option(
             "--header/--no-header",
--- a/pom.xml
+++ b/pom.xml
@@ -986,7 +986,7 @@
                   <exclude>**/src/license/**</exclude>
                   <exclude>**/target/**</exclude>
                   <exclude>**/node_modules/**</exclude>
-                  <exclude>**/.venv/**</exclude>
+                  <exclude>**/.venv*/**</exclude>
                   <exclude>**/spark-warehouse/**</exclude>
                   <exclude>**/metastore_db/**</exclude>
                   <exclude>**/.ipynb_checkpoints/**</exclude>
--- a/site/docs/libraries/cli.md
+++ b/site/docs/libraries/cli.md
@@ -70,7 +70,7 @@ as a human-readable table by default.
 | Option                           | Behaviour                                                                                                                     |
 | -------------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
 | `--format`                       | `table` (default), `csv`, `ndjson`; with `-o` also `parquet`, `delta`.                                                        |
-| `-o PATH`                        | Write to a file instead of stdout; the format is inferred from the extension (`.csv`/`.tsv`, `.ndjson`/`.jsonl`, `.parquet`). |
+| `-o PATH`                        | Write to a file instead of stdout; the format is inferred from the extension (`.csv`/`.tsv`, `.ndjson`/`.jsonl`, `.parquet`). A `.tsv` extension selects the CSV format only - it does not set the delimiter. |
 | `--limit N`                      | Row cap for stdout table output (default 50).                                                                                 |
 | `--overwrite`                    | Allow replacing an existing output path.                                                                                      |
 | `--departition/--no-departition` | Write file output as a single file (default) or as a Spark directory of part files. No effect on Delta.                       |
@@ -88,7 +88,10 @@ full result.
 The `--delimiter` and `--header/--no-header` options apply to CSV output for
 `view`, `fhirpath`, and the terminology commands, and are ignored for non-CSV
 formats. The delimiter is also used to read CSV input in the terminology
-commands, so a tab-separated dataset round-trips in a single invocation:
+commands, so a tab-separated dataset round-trips in a single invocation. Note
+that a `.tsv` extension only selects the CSV format - it does not imply a tab
+delimiter, so `--delimiter '\t'` must still be given explicitly on both the
+input and the output side:
 
 ```bash
 pathling member-of codes.tsv --code-column code \
@@ -296,10 +299,12 @@ The default result column names (`member_of`, `translated_system` and
 `designation`) can be overridden with `--result-column`.
 
 CSV input is read with the shared `--delimiter` (so a tab- or semicolon-separated
-dataset is read correctly); a `.tsv` file is read as CSV. Pass `--no-input-header`
-to read a headerless CSV; its columns are then addressable by the positional names
-Spark assigns (`_c0`, `_c1`, ...), which you reference via `--code-column`,
-`--system-column`, and the other column options.
+dataset is read correctly); a `.tsv` file is read as CSV, but the extension
+alone does not select a tab delimiter - `--delimiter '\t'` must still be given
+explicitly. Pass `--no-input-header` to read a headerless CSV; its columns are
+then addressable by the positional names Spark assigns (`_c0`, `_c1`, ...),
+which you reference via `--code-column`, `--system-column`, and the other
+column options.
 
 ```bash
 pathling member-of headerless.csv --no-input-header --code-column _c0 \
--- a/site/docs/libraries/cli.md
+++ b/site/docs/libraries/cli.md
@@ -185,13 +185,23 @@ The Pathling public functions are also pre-imported, so no
 coding functions (`to_coding`, `to_snomed_coding`, `to_loinc_coding`,
 `member_of`, `translate`, `subsumes`, and so on), the argument helper types
 (`Coding`, `PropertyType`, `Equivalence`), and the API types (`PathlingContext`,
-`DataSource`, and the rest). See the
-[Python API reference](https://pathling.csiro.au/docs/python/pathling.html) for
-the full list. The terminology
+`DataSource`, and the rest). The pre-imported names are exactly the
+[public functions and types listed in the Python API
+reference](https://pathling.csiro.au/docs/python/pathling.html#module-pathling).
+The terminology
 display function is bound under both its natural name `display` and the alias
 `tx_display`. Any pre-imported name is only a default: assigning or defining the
 same name in your own code takes precedence, and an explicit
 `from pathling import ...` continues to work unchanged.
+
+:::warning
+The `pathling` name is bound to the configured context, not the `pathling`
+module, so a bare `import pathling` inside a script or at the console
+replaces that variable with the module and breaks any later
+`pathling.read(...)`/`pathling.view(...)` call. This does not affect
+`from pathling import ...`, which only binds the names it imports and leaves
+`pathling` itself untouched.
+:::
 
 ```bash
 # Run a script file.


piotrszul:
Feedback on #2655 (Add a CSV delimiter option to the CLI):

## Design: `.tsv` should be its own format default, not just a CSV-format alias

Right now `.tsv` only affects format *inference* (`-o out.tsv` → `--format csv`) — the delimiter is never inferred from the extension; `--delimiter '\t'` must be passed explicitly on both read and write, even for a `.tsv` path.

A quick survey of how other tools handle this:
- **Spark** has no distinct TSV format — it's CSV with a `sep`/`delimiter` option, defaulting to `,` regardless of extension. This is the model the CLI currently mirrors.
- **pandas** has no `.tsv`-extension inference either, but it does give tab its own default via a *separate function* — `read_csv` (`sep=','`) vs. `read_table` (`sep='\t'`).
- **R** follows the same pattern on read (`read.csv` vs. `read.delim`, `,` vs. `\t`), though asymmetrically — there's no `write.delim`; tab output requires `write.table(x, sep = "\t")`.
- **Excel** is the outlier: it has no native "Save as TSV" (tab output is always `.txt`), but on *read*, it does auto-detect `.tsv` and split columns on tab, while a same-content `.txt` opens as one unparsed column.

None of these infer the delimiter purely from `.tsv` the way a filesystem "magic extension" would, but pandas/R both establish precedent for **`,` and `\t` as independent per-format defaults** rather than one option with a single global default. I'd suggest treating `.tsv` as a first-class format in the CLI — auto-detected from the extension on both read and write, defaulting `--delimiter` to `\t` for it (`,` for `.csv`) — while `--delimiter` remains available to override either. That removes the current footgun (silently comma-delimiting a `.tsv` output, or misreading a `.tsv` input, if `--delimiter` is forgotten) without giving up explicit control.

## Technical: missing type annotations in the CLI implementation

The delimiter/header plumbing added in #2655 leaves several new/touched signatures without type hints in `lib/python/pathling/cli/render.py` and `lib/python/pathling/cli/terminology.py` — e.g. `_delimiter_callback`'s `ctx`/`param`, `_read_dataset`'s `delimiter`/`input_header`/`from_format`, and the `delimiter`/`header`/`input_header` parameters threaded through `_execute`, `_run_subsumption`, and the terminology subcommands. `CONTRIBUTING.md`'s Python conventions call for type hints on all function parameters and return values. Worth a follow-up pass, particularly since these are the extension points other commands compose against.

---
*Moved from #2656.*

piotrszul:
Feedback on #2654 (Pre-import Pathling functions into the run and console namespace):

## Design: the `pathling` context variable shadows the `pathling` module

`run` and `console` bind the configured `PathlingContext` to the name `pathling` (`namespace["pathling"] = pc` in `run.py`; `user_ns["pathling"] = pc` in `console.py`). That's the same name as the package itself, so a bare `import pathling` inside a script, or typed at the console, silently rebinds `pathling` to the module object and clobbers the context - any later `pathling.read(...)`/`pathling.view(...)` call then fails.

Why this is worth fixing rather than just documenting (we've added a `:::warning` admonition to `cli.md` for it in the meantime):

- **The failure is silent and deferred.** Nothing errors at the `import pathling` line itself - the break only surfaces later as a confusing `AttributeError: module 'pathling' has no attribute 'read'`, disconnected from its actual cause.
- **`import pathling` is exactly the reflex this feature is supposed to make optional, not dangerous.** Anyone coming from the library API (where the normal pattern is `import pathling; pc = pathling.PathlingContext.create(...)`), or just following ordinary Python habit/IDE auto-import, is likely to type it - and in the CLI it actively breaks their script instead of being a redundant no-op.
- **Reusing a top-level package's own name for a differently-typed runtime object in the same scope is a general anti-pattern** (the same reason nobody names a variable `os` for something that isn't the `os` module) - it's a shadowing footgun independent of this specific feature.
- **Now is the cheapest time to change it.** 9.9.0 hasn't been tagged or merged to `main` yet (last release is v9.8.0), so renaming the bound variable - e.g. to `pc`, `ctx`, or `px` - costs nothing in backward compatibility today, whereas it becomes a breaking change for every existing `run`/`console` script the moment this ships.

Suggest renaming the context variable to something that doesn't collide with the module name, and dropping the doc warning once the collision no longer exists.

---
*Moved from #2656.*

piotrszul:
Follow-up feedback on #2645 (local terminology evaluation), found while testing local-mode import against real-world RF2 packages — not a regression in this PR's diff, but worth tracking.

## 🟢 Enhancement requests

**1. Silent, unreported data loss when a source package isn't self-contained** ✅ — `terminology/src/main/java/au/csiro/pathling/terminology/store/SnomedRf2Importer.java`
`readDescriptions`/`readRelationships`/`readRefsets` resolve every row via an inner join against `denseByCode`, built solely from the current import's own Concept file. Rows referencing a concept absent from that same source are dropped silently — no log line, no warning, exit code 0. This is the normal shape of any *derived* SNOMED package (one that declares a dependency via the Module Dependency reference set) shipped without its dependency's content:
- SNOMED IPS (0 native concepts, all-external refset) throws `"The release contains no active concepts..."` — the one existing guard, which only fires on total emptiness.
- SNOMED CT Netherlands Patient Friendly Extension (2 native bookkeeping concepts, 1,287 active descriptions): only 7 resolve; the other 1,280 (~99.5%) target un-shipped International-core concepts. This import **succeeds**, logging `"Writing 2 concepts to the store"`, with nothing indicating 99.5% of its payload vanished.

→ Initial solution: document the requirement in `terminology.md`/`cli.md` — a derived/extension package needs its declared dependency's content merged into the source before import, or referencing content will be silently dropped. (A code-level fix — logging input-vs-resolved counts per table, and broadening the "no active concepts" guard into a general "most referenced content didn't resolve" warning — could follow as a separate enhancement.)

**2. Feature request: support merging multiple RF2 sources on import** — `import-snomed` CLI/API
Today `import-snomed` takes one source, with no built-in way to combine a derived package with its dependency's release. A manual merge of extracted directory trees works, but is fiddly and has a sharp edge: `locateFiles` treats `concept`/`relationship`/`moduleDependency` as single fields (last-match-wins on multiple matches), so naively merging two full trees can silently pick the wrong file.

→ Initial solution: document the manual-merge recipe as a supported workaround — which files to combine, which to leave out, and the last-match-wins pitfall. (Accepting a list of sources and merging them internally — ideally validating the supplied dependency's module/version against the primary package's own Module Dependency refset — could follow as a separate feature.)

## Assessment
Both are documentation-first proposals — low-risk initial fixes with code-level follow-ups noted for later if there's appetite.

<sub>🤖 Found via interactive testing with Claude Code, tracing `SnomedRf2Importer` against real SNOMED IPS/UK Monolith/Netherlands RF2 packages.</sub>

piotrszul:
Follow-up feedback on #2645 (local terminology evaluation) — an improvement proposal for the local store's memory footprint. Not a regression in this PR's diff; raised while tracing the in-memory index structures.

## 🟢 Improvement proposal: assign dense IDs by hierarchy locality to shrink the closure bitmaps

**Where** — `terminology/src/main/java/au/csiro/pathling/terminology/store/SnomedRf2Importer.java`
Dense concept IDs are assigned by code order: `row_number().over(Window.orderBy(COLUMN_CODE)).minus(1)`. Those IDs address every runtime `RoaringBitmap`, and the largest consumer is `HierarchyIndex`, which holds the full transitive closure **twice** (`descendants` + `ancestors`) as `Map<Integer, RoaringBitmap>`.

**Why it costs memory** — A RoaringBitmap stores each 65 536-value chunk as an *array* container (~2 bytes/value), a *bitset* container (fixed 8 KB), or a *run* container (~4 bytes/run, independent of length). A concept's descendant set is a subtree of the is-a hierarchy, but an SCTID's numeric value is uncorrelated with hierarchy position — so under code-order the subtree lands as a random scatter across `[0, N)`: no runs, many small array containers, large sets tipping into 8 KB bitsets. E.g. *Clinical finding*'s ~180 k descendants span the whole ID range as several array/bitset chunks (tens of KB), repeated for every internal node.

**Proposal** — Assign dense IDs by a DFS pre-order / interval labelling over the active is-a graph at import time, leaving everything downstream unchanged. Each subtree then occupies a near-contiguous ID interval, so descendant sets collapse to one or a few *run* containers — an order-of-magnitude reduction on the map that dominates the footprint. `subsumes`/`member_of` also speed up (fewer, longer runs to AND/`contains`).

Caveats worth stating up front:
- SNOMED is a DAG, so one linear order can't make both directions contiguous. DFS pre-order makes **descendant** sets contiguous (the large, expensive ones) and leaves **ancestor** sets scattered (already cheap — bounded by depth). That's the right trade: optimise the direction that costs memory.
- Code-order presumably buys **deterministic, stable IDs across re-imports** (only new/retired codes shift). A DFS order is still deterministic per input but reshuffles more between releases — a genuine compactness-vs-stability trade-off, not a free win.

→ Initial step: quantify it. The benchmark added in this PR (`benchmark/…/TerminologyBenchmarkState`) can load a full International edition under both orderings and compare `HierarchyIndex` heap; I'd expect `descendants` to shrink sharply and `ancestors` to stay flat. If the measured win holds, gate the ordering behind an import option so the determinism trade-off stays opt-in.

**Update — empirical baseline under the current (code-order) scheme:** ran a smoke test exercising all seven terminology UDFs (`member_of`, `translate`, `subsumes`, `subsumed_by`, `display`, `property_of`, `designation`) in local mode against a store imported from the full UK SNOMED CT release (`uk_sct2mo_39.2.0`, edition `999000021000000109`, 1,115,237 concepts — International content plus the UK extension, the largest single edition on hand). Binary-searched `spark.driver.memory` for the minimum that avoids OOM, at 0.5 GB precision:

- **1 GB (Spark's local-mode default): OOMs** in `MemberOfUdf`, building `HierarchyIndex`.
- **3.5 GB: OOMs** in `MemberOfUdf`/`DesignationUdf`, building the dictionary/hierarchy indexes.
- **4 GB: passes** cleanly across all seven UDFs.

So under today's code-order dense IDs, a full national-extension edition needs **~4 GB of driver heap** just to hold the closure/hierarchy indexes for ordinary terminology operations — a concrete number for the DFS pre-order proposal above to be measured against.

## Assessment
A measurement-first proposal targeting the local store's dominant in-memory cost. No behavioural change — dense IDs are an internal addressing detail — so the only risk is import-time reproducibility, which the caveats above bound.

<sub>🤖 Found via interactive analysis with Claude Code, tracing the local terminology index structures (`ConceptDictionary`/`HierarchyIndex`) and RoaringBitmap container behaviour in #2645. Memory baseline measured via interactive testing with Claude Code against the local store built from the UK SNOMED CT RF2 release.</sub>

piotrszul:
## Misc, other etc

Unrelated to this PR's diff, but discovered along the way while working through local terminology testing:

- **#2669** — the CLI's `--spark-conf` has no equivalent on the Python `PathlingContext.create()` API (only a `spark=` pre-built-session escape hatch), and the `spark` parameter's docstring is stale (references a bundled JAR and a `pathling.etc.find_jar` helper that no longer exist now that Pathling resolves via Maven coordinates).

<sub>🤖 Found via interactive testing with Claude Code.</sub>

johngrimes:
I've measured this, and the proposal holds up - though for a different reason than either of us expected, and with a couple of the stated predictions coming out the other way.

Two things I found before taking any numbers changed the shape of the measurement. First, the bitmap library only produces run-length-encoded chunks when explicitly asked to, and the index never asks, so there are no such chunks today at any ordering - which makes `runOptimize()` an independent factor worth measuring in its own right rather than an implementation detail. Second, RoaringBitmap's own `getLongSizeInBytes()` documents itself as potentially 10x out for "a small set of random-looking integers spanning a wide range of values", which is exactly the status quo being baselined, so it can't be the instrument. I used JOL's graph layout for true retained heap instead.

So there are four variants, crossing identifier ordering with representation optimisation: A code order unoptimised (today), B code order optimised, C pre-order unoptimised, D pre-order optimised.

## Results

SNOMED CT International 20210131, 481,509 concepts, retained heap in bytes:

| Variant | Total | vs A | vs B | vs C |
| --- | --- | --- | --- | --- |
| A code order, unoptimised | 309,577,520 | baseline | | |
| B code order, optimised | 308,460,592 | -0.4% | | |
| C pre-order, unoptimised | 238,329,448 | -23.0% | | |
| D pre-order, optimised | 229,692,096 | -25.8% | -25.5% | -3.6% |

SNOMED CT UK 39.2.0, 1,115,237 concepts:

| Variant | Total | vs A | vs B | vs C |
| --- | --- | --- | --- | --- |
| A code order, unoptimised | 738,063,976 | baseline | | |
| B code order, optimised | 737,968,480 | -0.0% | | |
| C pre-order, unoptimised | 548,134,752 | -25.7% | | |
| D pre-order, optimised | 536,177,136 | -27.4% | -27.3% | -2.2% |

Repeat runs of a variant agree exactly - the spread is 0.00% on all four variants on both editions, because the construction and the instrument are both deterministic. That matters here, since it means a 2% difference is a result rather than noise.

## Three predictions that didn't hold

**The descendant maps don't dominate; the ancestor maps do.** On the UK edition variant A is descendants 80.6 MB, ancestors 413.2 MB, children 54.2 MB, parents 190.1 MB. Ancestors are 56% of the total on their own. The premise that the descendant maps are the cost is wrong.

**The ancestor maps don't stay flat.** They shrink 35.4% under the pre-order, more than the descendant maps do in absolute terms (146 MB against 36 MB). The directed-acyclic-graph objection was that one linear order can't make both directions contiguous, and that's true, but the ancestor sets still benefit: a concept's ancestors are largely its own chain, and a pre-order clusters that chain.

**Variant C isn't meaningless without `runOptimize()`.** It was expected to be a no-op, since without run-length encoding there'd be nothing to gain. In fact it delivers almost the whole saving. The container histogram says why: the pre-order takes the UK index from 7,605,623 array containers to 4,228,733. The mechanism isn't run-length encoding at all, it's that contiguity puts each set into far fewer 16-bit chunks, and every chunk carries a container object and its array. Run-length encoding then adds a further 2.2%.

The corollary is that `runOptimize()` alone is not the cheap answer: it saves 0.4% on the International edition and 0.01% on the UK one. It's worth having on top of the reordering, not instead of it.

## Driver heap

The claim being tested was around 4 GB for a full national edition. Running all seven terminology functions over the UK store through the CLI in 0.5 GB steps:

| | Minimum viable driver memory |
| --- | --- |
| Status quo | 4.0 GB |
| Pre-order | 3.5 GB |
| Reported baseline | 4 GB - reproduced exactly |

Worth noting what's actually binding at the bottom: below 3.5 GB it's `designation` that fails first, on the description index, not the hierarchy index. So the hierarchy index isn't the only thing standing between this and a smaller heap.

## Decision

The rule was written into the plan before any number was taken: adopt only if the reordering reduces total retained heap by at least 25% beyond what `runOptimize()` achieves alone, or lowers the UK edition's minimum viable driver heap by at least one 0.5 GB step. Both are met (-27.3% D against B; 4.0 GB to 3.5 GB), so the reordering is adopted, behind an option and off by default.

One thing the measurement surfaced that's worth recording: making the change genuinely invisible took more than the reordering itself. Multi-valued lookup properties - a concept's parents, children, and its defining-relationship targets - were being emitted in internal identifier order, so changing that order changed what `property_of` returned. Over the UK edition, Diabetes came back with its parents as `126877002, 362969004` under one ordering and reversed under the other. They're now ordered by concept code, which reproduces today's output exactly and makes the internal choice unobservable. All seven functions are byte-identical across the two orderings over the full UK edition, including a property lookup over concepts with 29 same-type targets.

johngrimes (approved):
Thanks @piotrszul, I have addressed these items in #2673, #2674 and #2675.

Issue #2594, opened 2026-04-24
trace() entries are duplicated by non-deterministic expression re-evaluation


piotrszul:
## Summary

`trace()` entries captured by `PathlingContext.evaluateFhirPath()` are emitted more than once when the traced column is consumed by any operation that compiles into a `when(cond(c), …).otherwise(expr(c))` pattern in `ColumnRepresentation`. The collector receives 2× (or more) entries per logical trace firing, causing the FHIRPath Lab API response (and any SLF4J log output) to contain spurious duplicates.

## Reproduction

Input Patient with three names:

```json
{
  "resourceType": "Patient",
  "name": [
    {"use": "official", "family": "Smith", "given": ["John", "Quincy"]},
    {"use": "usual",    "family": "Smith", "given": ["Johnny"]},
    {"use": "maiden",   "family": "Doe",   "given": ["John", "Q"]}
  ]
}
```

Using `pc.evaluate_fhirpath("Patient", patient_json, expr)` (Python) or the equivalent Java API:

| Expression | Expected trace entries | Actual |
|---|---|---|
| `name.trace('trc')` | 3 | 3 ✓ |
| `name.trace('trc').given.join(' ')` | 3 | 3 ✓ |
| `name.trace('trc').given.join(' ') + 'X'` | 3 | 3 ✓ |
| `name.trace('trc').given.count()` | 3 | **6** |
| `name.trace('trc').exists()` / `empty()` | 3 | **6** |
| `name.trace('trc').first()` / `last()` | 3 | **6** |
| `name.trace('trc').given.join(' ').combine('X')` | 3 | **6** |
| `name.trace('trc').given.join(' ') \| name.family.first()` | 3 | **6** |
| `name.trace('trc') \| name.trace('trc')` | 6 | **12** |

## Analysis

`TraceExpression` (`encoders/src/main/scala/au/csiro/pathling/sql/TraceExpression.scala`) is declared `Nondeterministic`. This is intentional — it ensures Catalyst will not eliminate the expression or reorder it around side-effecting boundaries.

A side effect of `Nondeterministic` is that Catalyst's `CommonSubexpressionElimination` rule **excludes the expression from CSE**. Any pattern in the generated Spark plan that references the same operand column twice will therefore re-evaluate the entire subtree, firing the trace collector twice per logical input element.

Several widely-used operations in `ColumnRepresentation` are built on exactly this shape. Each references `c` two or three times:

[...]

Issue #437, opened 2022-02-24
Extension support for primitive elements


johngrimes:
Our current implementation of extension support does not include extensions on primitive elements.

This will add this support and assess any impacts on query performance.

johngrimes:
I feel as though we should deprioritise this until such time as we find a solid use case in need of it.

jasminziegler:
Dear @johngrimes, I am currently trying to work with extensions. Is this what you mean by "primitive elements"?
![image](https://github.com/aehrc/pathling/assets/29899146/5c9f2bf6-84cc-48dd-a99c-107151053505) (taken from https://simplifier.net/oncology/operation) 
I was expecting to be able to extract the value of the codeable concept, but the "extension_url" is as far as i can get (extension_enabled = True).

johngrimes:
Hi @jasminziegler,

Using this example resource: https://simplifier.net/oncology/Operation/~json

This expression works for me:

```
extension('http://dktk.dkfz.de/fhir/StructureDefinition/onco-core-Extension-OPIntention').valueCodeableConcept.coding.display
```

Returns one row with "palliative".

jasminziegler:
Hi @johngrimes, thanks for your quick reply!
I might have confused the pathling fhir-server implementation and the pathling python api - is this functionality also available in the pathling python api?

johngrimes:
As of this morning, yes! 🙂

Here is the newly minted documentation on how to do FHIRPath query using the library: https://pathling.csiro.au/docs/libraries/fhirpath-query

I'd love to hear any feedback you might have!

jasminziegler:
Awesome, thank you @johngrimes! 
We are waiting for the maven artifact and are ready for testing the new and exciting features :)

johngrimes:
You can use this one: https://central.sonatype.com/artifact/au.csiro.pathling/library-api/6.2.1

jasminziegler:
Got it upgraded and installed! 
Nevertheless, I am getting an AttributeError: 'DataFrame' object has no attribute 'extract'.
Not sure what I am missing here bec. according to this example https://pathling.csiro.au/docs/libraries/fhirpath-query, reading in data with "pc.read..." will produce a DataFrame.

 'PATHLING_VERSION': '6.2.1',
'APACHE_SPARK_VERSION': '3.3.2',
Python 3.10.10
Scala version 2.12.15

Could you please provide your example that you tested with my sample resource?

johngrimes:
[...]

PR #2705, opened 2026-08-06
feat: Support most existence functions (#2385)

### LINKED ISSUE #2385: Support all existence functions
Spec link: https://hl7.org/fhirpath/#existence

This includes the following functions:

- all
- subsetOf
- supersetOf
- isDistinct
- distinct
- count
- allTrue
- allFalse
- anyTrue
- anyFalse

piotrszul:
## Summary

- Implements 7 of the 9 FHIRPath existence functions from #2385: `all()`, `allTrue()`, `allFalse()`, `anyTrue()`, `anyFalse()`, `isDistinct()`, `distinct()`. `count()` was already implemented.
- `all(criteria)` is implemented as `where(criteria).count() = count()`, reusing `where()`'s existing "evaluates to true" semantics rather than re-deriving per-item truthiness.
- `allTrue()`/`allFalse()`/`anyTrue()`/`anyFalse()` delegate to pre-existing `ColumnRepresentation` helpers (already used elsewhere, e.g. by SQL on FHIR view constraints) that were not yet exposed as FHIRPath functions.
- `isDistinct()`/`distinct()` reuse the same equals-based deduplication as `union()`/`combine()` (`CombiningLogic.dedupeArray`) via a new package-private `ExistenceLogic` helper, and reject complex/non-equatable types with the same `UnsupportedFhirPathFeatureError` used elsewhere for that limitation.
- `all()` and `isDistinct()` materialise their input column once with `let()` before deriving the two counts they compare, avoiding the double-evaluation-of-nondeterministic-operands class of bug tracked under #2594 (caught by an independent review; regression cases added to `TraceFunctionTest`).

Resolves #2385.

**`subsetOf()`/`supersetOf()` are deliberately out of scope for this PR.** They are the second pair of functions (after `union()`/`combine()` in #2587) to need a plain `collection`-typed argument evaluated against the surrounding iteration focus rather than `%context`, and Pathling's standard function-invocation path (`FunctionParameterResolver`) doesn't support that — the same architectural gap #2587's design doc documented and deliberately deferred. Fixing it requires the same shape of work as #2587 (a new operator plus a parser desugaring rule), which needs its own OpenSpec change. Tracked as follow-up in #2706.

## Verification

- `ExistenceFunctionsDslTest` extended with 135 assertions across empty/singleton/array/computed-empty cardinality, `$index` usage, non-Boolean-input errors, the full equatable type matrix (Boolean, Integer, Decimal, String, Date, DateTime, Time, Quantity, Coding) for `isDistinct()`/`distinct()`, and complex-type errors.
- `TraceFunctionTest$TraceEntryCountTest` extended with 2 regression cases guarding the #2594 fix in `all()`/`isDistinct()`.
- `fhirpath/src/test/resources/fhirpath-js/config.yaml` (`5.1_existence.yaml`): 6 pre-existing reference-corpus divergences newly exposed by this feature were added to the exclusion baseline, each extending an existing documented limitation in this same file (null-value representation in boolean arrays, #437 primitive-extension-without-value, indefinite calendar duration equality) rather than a new category.
- Full `fhirpath` module: 6784 tests, 0 failures, 0 errors.

### DIFF
--- a/fhirpath/src/main/java/au/csiro/pathling/fhirpath/function/provider/ExistenceFunctions.java
+++ b/fhirpath/src/main/java/au/csiro/pathling/fhirpath/function/provider/ExistenceFunctions.java
@@ -17,6 +17,7 @@
 
 package au.csiro.pathling.fhirpath.function.provider;
 
+import static au.csiro.pathling.sql.SqlFunctions.let;
 import static java.util.Objects.nonNull;
 
 import au.csiro.pathling.fhirpath.annotations.SqlOnFhirConformance;
@@ -24,10 +25,13 @@ import au.csiro.pathling.fhirpath.annotations.SqlOnFhirConformance.Profile;
 import au.csiro.pathling.fhirpath.collection.BooleanCollection;
 import au.csiro.pathling.fhirpath.collection.Collection;
 import au.csiro.pathling.fhirpath.collection.IntegerCollection;
+import au.csiro.pathling.fhirpath.column.ColumnRepresentation;
+import au.csiro.pathling.fhirpath.column.DefaultRepresentation;
 import au.csiro.pathling.fhirpath.function.CollectionTransform;
 import au.csiro.pathling.fhirpath.function.FhirPathFunction;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
+import org.apache.spark.sql.functions;
 
 /**
  * Contains functions for evaluating the existence of elements in a collection.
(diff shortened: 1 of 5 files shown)


piotrszul:
### Code review

Found 1 issue:

1. The new `distinct()` exclusion matches on a case *description* that fhirpath.js reuses across four case files, so it shadows two existing, more specific exclusion rules. Matching is expression-**or**-description ([`AnyPredicate`](https://github.com/aehrc/pathling/blob/abbf825f36eb1b3c6f33646a4bf95db3f9404eb7/fhirpath/src/test/java/au/csiro/pathling/test/yaml/format/AnyPredicate.java#L33-L37)), all blocks are flattened in document order and the first match wins ([`YamlTestFormat.toExcluder`](https://github.com/aehrc/pathling/blob/abbf825f36eb1b3c6f33646a4bf95db3f9404eb7/fhirpath/src/test/java/au/csiro/pathling/test/yaml/format/YamlTestFormat.java#L55-L61)), and an exclude block's `glob` is inert so every block applies to every case file. Sitting at line 177, this rule therefore intercepts the `repeat($this)` case (`5.2_filtering_and_projection.yaml:90`) and the `(1 year | 12 months)` case (`5.4_combining.yaml:23`) ahead of their own rules.

https://github.com/aehrc/pathling/blob/abbf825f36eb1b3c6f33646a4bf95db3f9404eb7/fhirpath/src/test/resources/fhirpath-js/config.yaml#L176-L188

The build stays green because all three rules share `outcome: failure`, so this is a baseline-integrity problem rather than a breakage: both cases are now logged as excluded under the title "Indefinite calendar duration equality in **distinct()** deduplication", and these two rules become unreachable for the cases they were written for.

https://github.com/aehrc/pathling/blob/abbf825f36eb1b3c6f33646a4bf95db3f9404eb7/fhirpath/src/test/resources/fhirpath-js/config.yaml#L231-L233

https://github.com/aehrc/pathling/blob/abbf825f36eb1b3c6f33646a4bf95db3f9404eb7/fhirpath/src/test/resources/fhirpath-js/config.yaml#L532-L534

Suggested fix: match on the expression instead, which is unique to the intended case, consistent with the two rules it currently shadows:

```yaml
        any:
          - "(1 year).combine(12 months).distinct()"
```

The other four description matchers added by this PR (the `allTrue`/`anyTrue`/`allFalse`/`anyFalse` "raise an error" titles) are unique to `5.1_existence.yaml` and do not have this problem.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

piotrszul:
Confirmed and fixed in 0a83a92338: the exclusion now matches on the expression `(1 year).combine(12 months).distinct()` instead of the shared description string. Verified the `repeat($this)` and `(1 year | 12 months)` exclusions it was shadowing are unaffected — `YamlReferenceImplTest` is still 0 failures, 0 errors (1821 run, 920 skipped as excluded).

johngrimes (approved):
Looks good to me!

Thanks @piotrszul!

Issue #2724, opened 2026-08-11
$sql-export does not honour its documented kick-off validation guarantee


fhnaumann:
**Describe the bug**

`sql-export.md` states the following guarantee for `$sql-export`:

> **Validated at kick-off.** Every subject is resolved, named and prepared before the job starts,
> and every problem in the request is reported in one `OperationOutcome`. **A job that starts is
> never rejected later at its status URL.**

That last sentence does not hold for a whole class of problem. Kick-off runs
`SqlExportRequestParser` → `SqlQueryPipeline.validateStatically`, which delegates to `SqlValidator`
only. `SqlValidator` works on the parsed logical plan and never triggers Spark's analyser, so any
fault the analyser is the one to catch — an unresolved column, an unknown function, a missing
`GROUP BY`, an ambiguous reference, a datatype mismatch — sails through kick-off, is accepted with a
`202`, and only fails once the job thread runs the query.

The caller therefore receives a `202` and a status URL for a request that was already known to be
unrunnable at the moment it was accepted, which is exactly the situation the documented guarantee
promises cannot happen.

**To Reproduce**

`POST /fhir/$sql-export` with a SQLQuery `Library` subject whose SQL names a column that does not
exist. Kick-off returns `202` with a `Content-Location`; polling that URL eventually reports a
failure.

**Expected behavior**

Either the guarantee holds — kick-off analyses each subject and rejects an unanalysable one
synchronously — or the documented guarantee is narrowed to say precisely which classes of problem
it covers.

**Additional context**

This is not a small fix, which is why it is filed separately rather than folded into #2723.
Triggering analysis at kick-off requires the resolved dependency graph to be materialised as temp
views so the analyser has something to resolve against, then dropped, then materialised again on the
job thread — or a refactor that lets the two share one materialisation. `SqlQueryPipeline.execute`
currently owns that whole lifecycle, and the kick-off thread and the job thread are different
threads with different lifetimes.

Note that #2723 makes the deferred failure at least *legible* — a `422` with Spark's own analyser
message rather than an opaque `500` — but it does not move the detection earlier, so this defect
survives that fix.

johngrimes:
I don't think this documented guarantee is correct, or possible.

I think the fix for this is to update the documentation.

PR #2726, opened 2026-08-11
fix: Report a Spark analysis failure as a 422 naming the subject

### LINKED ISSUE #2723: Spark analysis errors on $sql-run and $sql-export are reported as an opaque 500
**Describe the bug**

Every Spark *analysis* failure raised by a `$sql-run` or `$sql-export` subject is reported as an
opaque `500 Internal Server Error` with the body:

```json
{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"processing","diagnostics":"Unexpected error occurred"}]}
```

`org.apache.spark.sql.AnalysisException` extends `java.lang.Exception` and matches no clause in
`ErrorHandlingInterceptor.convertError`, so it falls through to the terminal
`catch (Throwable) -> internalServerError` branch, which discards the cause. The real message is
produced, and reaches the container log, but never the client:

```
Caused by: org.apache.spark.sql.AnalysisException: [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column,
  variable, or function parameter with name `no_such_col` cannot be resolved. Did you mean one of
  the following? [`bd`, `gender`, `pid`]. SQLSTATE: 42703; line 1 pos 7;
```

The consequence is that a *client* mistake — a typo in a column name in a submitted SQLQuery
Library — is indistinguishable, both by status class and by body, from a genuine server-side or
infrastructure fault. A client that (reasonably) treats `4xx` as "my request was wrong, fix it and
retry" and `5xx` as "the server is unhealthy, do not retry" gets exactly the wrong answer for the
single most common class of SQL mistake.

This also contradicts the design intent already recorded in the code. `SqlValidator.validateFunctionName`
(`server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlValidator.java:781-788`) deliberately
declines to reject unknown function names so that *"Spark's analyser produces its standard 'function
not found' error, which is more helpful than a synthetic rejection"*. That message is then thrown away
by the error handler, so the intended behaviour is not what a caller observes.

**To Reproduce**

Against a server with data loaded (auth off), `POST /fhir/$sql-run` with an inline SQLQuery `Library`
subject whose SQL names a column that does not exist, e.g. `SELECT no_such_col FROM t`.

Observed status matrix on `3.0.0-SNAPSHOT+378dba8`. Every `500` below returned the byte-identical
body quoted above.

| SQL in the subject | Observed | Expected |
|---|---|---|
| `SELECT no_such_col FROM t` | 500 | 4xx with the Spark message |
| `SELECT * FROM (SELECT no_such_col FROM t) x LIMIT 0` | 500 | 4xx |
| `SELECT no_such_fn(pid) FROM t` | 500 | 4xx |
| `SELECT pid FROM t GROUP BY nonexistent` | 500 | 4xx |
| `SELECT pid, count(*) FROM t` (missing `GROUP BY`) | 500 | 4xx |
| `SELECT pid FROM t a, t b` (ambiguous reference) | 500 | 4xx |
| `SELECT pid[1] FROM t` (datatype mismatch) | 500 | 4xx |
| `SELECT 1/0 FROM t`, `CAST(pid AS INT)` (runtime, not analysis) | 500 | 500 — correct, see below |
| `SELEKT pid FROM t` | 400 | 400 — correct today |
| `SELECT x FROM nope` (undeclared table) | 400 | 400 — correct today |
| `DROP TABLE t` (disallowed operation) | 400 | 400 — correct today |

The three rows that are already `400` are the ones `SqlValidator` catches statically. Everything
Spark's analyser catches instead is a `500`.

**Expected behavior**

An analysis failure in a subject should be `422 Unprocessable Entity` with `issue.code = invalid`
and `expression = ["subject"]`, carrying Spark's own analyser message as `diagnostics`. This is
exactly what `sql-run.md`'s status table already defines `422` to mean: *"The subject is of no
admitted kind, **or is conformant but cannot be processed**"*. An unresolved column is a conformant
SQLQuery Library that cannot be processed. `400` stays reserved for the parameter-level faults that
table lists.

Notes on scope, for whoever picks this up:

- The translation belongs inside `operations/sql/`, not in `ErrorHandlingInterceptor`.
  `DeltaAnalysisException extends AnalysisException`, so a blanket catch in the interceptor would
  have to sit after the existing Delta carve-out and would silently widen behaviour across the
  import and Delta paths, where an analysis failure genuinely *is* a server-side fault.
- For `$sql-run`, the dataset is analysed inside `SqlQueryPipeline.execute` before the streaming
  consumer writes a byte, so the response is still uncommitted and the status is changeable.
- Runtime (non-analysis) errors — `1/0`, a failing `CAST` — fire inside the streaming consumer,
  after the response may already be committed, so they cannot be reclassified and should stay `500`.
- `operations/sql/` already has the pieces: `SqlOperationError.unprocessable(...)` and
  `SubjectResolver.SUBJECT_EXPRESSION`, and `SqlExportExecutor.runView` already demonstrates the
  pattern for `ConstraintViolationException`.

fhnaumann:
Fixes #2723.

## Problem

Every Spark *analysis* failure raised by a `$sql-run` or `$sql-export` subject reached the caller as
an opaque `500`:

```json
{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"processing","diagnostics":"Unexpected error occurred"}]}
```

`AnalysisException` extends `java.lang.Exception` and matches no clause in
`ErrorHandlingInterceptor.convertError`, so it fell through to the terminal
`catch (Throwable) -> internalServerError` branch, which discards the cause. An unresolved column,
an unknown function, a missing `GROUP BY`, an ambiguous reference and a datatype mismatch were all
byte-identical to a genuine infrastructure fault. `SqlValidator.validateFunctionName` deliberately
leaves unknown function names for Spark's analyser on the grounds that its message "is more helpful
than a synthetic rejection" — but that message never reached anyone.

## Change

An analysis failure is now translated inside `operations/sql/` into a `422` carrying
`issue.code = invalid`, `expression = ["subject"]` and Spark's own analyser message. That is what
`sql-run.md`'s status table already defines `422` to mean: *"The subject is of no admitted kind, or
is conformant but cannot be processed."* An unresolved column is a conformant SQLQuery `Library`
that cannot be processed, and `400` stays reserved for the parameter-level faults that table lists.

Three files, one new helper and two (well, three) call sites:

- `SqlOperationError.executionFailure(Exception)` — returns a `422` via the existing
  `unprocessable(...)` helper for an `AnalysisException`, and returns anything else unaltered so it
  still renders as a `500`. The diagnostics come from `getSimpleMessage()` rather than
  `getMessage()`: the latter appends the whole unresolved logical plan, which is unbounded, names
  the internal request-scoped views the dependency graph was materialised under, and tells the
  caller nothing they can act on. The result is length-bounded as well, since the "did you mean"
  list is drawn from the subject's columns and a wide dependency makes it long.
- `SqlRunProvider.runSqlLibrary` — wraps `pipeline.execute(...)`. The dataset is analysed inside
  `execute` before the streaming consumer writes a byte, so the response is uncommitted and the
  status is still free to change.
- `SqlExportExecutor.runSql` and `runView` — the same, beside the `ConstraintViolationException`
  catch that already demonstrates the pattern.

`AnalysisException` is a Scala-declared checked exception that no signature on the call path
declares, so it cannot be named in a `catch` clause. It is matched by type instead, following the
precedent in `ExportFileWriter.writeCsv`.

## Why not `ErrorHandlingInterceptor`

`DeltaAnalysisException extends AnalysisException`, so a blanket catch in the interceptor would have
to sit after the existing Delta carve-out and would silently widen behaviour across the import and
Delta paths, where an analysis failure genuinely *is* a server-side fault. The translation is
deliberately confined to `operations/sql/`.

## Why `$sql-export` needs no `JobProvider` change

`JobProvider` renders a failed job with `ErrorHandlingInterceptor.convertError(...)`, the same
converter as a synchronous request, and its `catch (BaseServerResponseException)` branch passes a
typed exception through unaltered when its status is non-zero. `SqlOperationError` produces an
`UnprocessableEntityException`, so a `422` thrown on the job thread survives the async round-trip
with its status and `OperationOutcome` intact.

## What deliberately does not change

- **Runtime (non-analysis) errors stay `500`.** `1/0`, a failing `CAST` and friends fire inside the
  streaming consumer, after the response may already be committed, so the status cannot be
  rewritten. Asserted negatively in the tests.
- **The three statically-detected faults stay `400`** — malformed SQL, an undeclared table, a
  disallowed operation. Asserted in `SqlRunProviderIT`.
- **Kick-off validation for `$sql-export` is unchanged.** An analysis error is still accepted with a
  `202` and surfaces as a deferred failure; this PR only makes that failure legible. Filed
  separately as #2724.

## Testing

`SqlRunProviderTest` — an analysis failure becomes a `422` naming the subject with the message
intact; a non-analysis failure propagates untranslated and is not a `BaseServerResponseException`.

`SqlExportExecutorTest` — the same pair, with `propagatesAFailingSubject` strengthened to assert the
untranslated propagation.

`SqlRunProviderIT` — real Spark analysis failures end to end over the wire: an unresolved column
(asserting the `UNRESOLVED_COLUMN` condition and the analyser's "did you mean" suggestions reach
`diagnostics`), an unresolved column nested in a subquery, an unknown function, and a regression
guard that the three statically-detected faults are still `400`.

```
SqlRunProviderTest, SqlExportExecutorTest    38 tests, 0 failures
SqlRunProviderIT, SqlExportProviderIT        29 tests, 0 failures
```

## Verified by hand

Against a locally built image, `POST /fhir/$sql-run` with an inline SQLQuery `Library` subject over
an inline `ViewDefinition` context projecting `pid`, `gender`, `bd`.

| SQL | Before | After |
|---|---|---|
| `SELECT no_such_col FROM t` | 500 opaque | **422** `[UNRESOLVED_COLUMN.WITH_SUGGESTION] ... Did you mean one of the following? [\`bd\`, \`gender\`, \`pid\`]. SQLSTATE: 42703; line 1 pos 7` |
| `SELECT * FROM (SELECT no_such_col FROM t) x LIMIT 0` | 500 opaque | **422** same condition, `line 1 pos 22` |
| `SELECT no_such_fn(pid) FROM t` | 500 opaque | **422** `[UNRESOLVED_ROUTINE] Cannot resolve routine \`no_such_fn\` ... SQLSTATE: 42883` |
| `SELECT pid FROM t GROUP BY nonexistent` | 500 opaque | **422** `[UNRESOLVED_COLUMN.WITH_SUGGESTION] ...` |
| `SELECT pid, count(*) FROM t` | 500 opaque | **422** `[MISSING_GROUP_BY] The query does not include a GROUP BY clause ...` |
| `SELECT pid FROM t a, t b` | 500 opaque | **422** `[AMBIGUOUS_REFERENCE] Reference \`pid\` is ambiguous, could be: [\`a\`.\`pid\`, \`b\`.\`pid\`]` |
| `SELECT pid[1] FROM t` | 500 opaque | **422** `[INVALID_EXTRACT_BASE_FIELD_TYPE] ...` |
| `SELEKT pid FROM t` | 400 | **400** unchanged, `PARSE_SYNTAX_ERROR` |
| `SELECT x FROM nope` | 400 | **400** unchanged, undeclared table |
| `DROP TABLE t` | 400 | **400** unchanged, disallowed operation |
| `SELECT 1/0 FROM t` | 500 | **500** unchanged — runtime, out of scope |
| `SELECT CAST(pid AS INT) FROM t` | 500 | **500** unchanged — runtime, out of scope |
| `SELECT pid, gender FROM t LIMIT 2` | 200 | **200** unchanged |

Every issue carries `"code":"invalid"` and `"expression":["subject"]`.


server/src/main/java/au/csiro/pathling/operations/sql/SqlOperationError.java
```diff
@@ -97,6 +103,51 @@ public static UnprocessableEntityException unprocessable(
...
+  @Nonnull
+  public static RuntimeException executionFailure(@Nonnull final Exception error) {
+    // AnalysisException is declared in Scala as a checked exception that no signature on the call
+    // path declares, so it cannot be named in a catch clause and is matched by type instead.
+    if (error instanceof final AnalysisException analysisError) {
+      return unprocessable(SubjectResolver.SUBJECT_EXPRESSION, analyserMessage(analysisError));
+    }
+    return error instanceof final RuntimeException runtimeError
+        ? runtimeError
+        : new RuntimeException(error);
```
johngrimes: Wrapping a checked exception in a plain `RuntimeException` breaks the interceptor's unwrap chain. The dominant checked exception here is `org.apache.spark.SparkException`, which is how every task-time failure inside the streaming consumer surfaces - and those consumers run inside the new try blocks (`SqlQueryExecutor.execute` invokes them before the temp views are dropped).

Pre-PR, a raw `SparkException` reached `ErrorHandlingInterceptor.convertError`, whose first branch recursively unwraps it and converts the cause. The plain `RuntimeException` wrapper matches none of the unwrap branches and falls through to the terminal `catch (Throwable)`, so these conversions regress to an opaque 500:

- FHIRPath singleton violations and `raise_error()`: `SparkRuntimeException` with `USER_RAISED_EXCEPTION` converts to a 400 carrying the message (`ErrorHandlingInterceptor:179-184`). `ErrorHandlingInterceptorTest.convertsUserRaisedExceptionTo400` shows the shape - "Expecting a collection with a single element but it has many" - which fires at streaming time in a ViewDefinition dependency of a SQL subject.
- `FhirClientConnectionException` -> 503 when a terminology call fails inside a task.
- The Delta schema-mismatch rendering at `ErrorHandlingInterceptor:191-202`.

`leavesANonAnalysisFailureUntranslated` only covers a `RuntimeException`, so this arm is untested.

Java's precise rethrow makes the arm unnecessary. Since `pipeline.execute` declares no checked exceptions, the call sites compile without a `throws` clause:

```java
} catch (final Exception e) {
  if (e instanceof final AnalysisException analysisError) {
    throw SqlOperationError.analysisFailure(analysisError);
  }
  // Precise rethrow: propagates byte-identically to the pre-PR behaviour.
  throw e;
}
```

(The cited precedent, `ExportFileWriter.writeCsv`, carries the same latent wrap - pre-existing and out of scope here, but worth not propagating.)

server/src/main/java/au/csiro/pathling/operations/sql/SqlOperationError.java
```diff
@@ -97,6 +103,51 @@ public static UnprocessableEntityException unprocessable(
...
+   *
+   * @param error the failure raised by the query engine
+   * @return the exception to throw
+   */
+  @Nonnull
+  public static RuntimeException executionFailure(@Nonnull final Exception error) {
+    // AnalysisException is declared in Scala as a checked exception that no signature on the call
+    // path declares, so it cannot be named in a catch clause and is matched by type instead.
+    if (error instanceof final AnalysisException analysisError) {
+      return unprocessable(SubjectResolver.SUBJECT_EXPRESSION, analyserMessage(analysisError));
```
johngrimes: `org.apache.spark.sql.delta.DeltaAnalysisException extends AnalysisException` (verified against delta-spark 4.0.0), so this branch translates Delta analysis failures into 422s blaming the subject - the very case the PR description rules out when arguing against an interceptor-level catch. Confining the translation to `operations/sql/` doesn't prevent it, because `pipeline.execute` materialises dependency views over the Delta source inside the wrapped call (`SqlQueryExecutor.materialiseNode`). Reachable cases:

- Schema drift the guard misses: previously the interceptor's dedicated rendering, which exists precisely because the raw Delta message embeds both struct definitions in full and cannot be returned (`ErrorHandlingInterceptor:191-202`). Now a 422 carrying up to 1024 characters of raw Delta message, including schema fragments.
- An `$sql-export` snapshot racing a vacuum: `org.apache.spark.sql.delta.VersionNotFoundException` (also `extends AnalysisException`) becomes a 422 telling the client their SQLQuery is invalid.

Suggest at minimum `!(error instanceof DeltaAnalysisException)` in this guard, mirroring the interceptor's carve-out. `VersionNotFoundException` remains a rare residual edge that is probably fine to note and leave.

server/src/main/java/au/csiro/pathling/operations/sql/SqlExportExecutor.java
```diff
@@ -156,6 +156,8 @@ private List<String> runView(
       throw SqlOperationError.unprocessable(
           SubjectResolver.SUBJECT_EXPRESSION,
           "The subject '%s' is invalid: %s".formatted(subject.name(), e.getMessage()));
+    } catch (final Exception e) {
+      throw SqlOperationError.executionFailure(e);
     }
```
johngrimes: This translation has no equivalent on the synchronous path: `$sql-run`'s `runViewDefinition` -> `streamView` is unwrapped, so the same ViewDefinition subject yields a 422 via `$sql-export` and an opaque 500 via `$sql-run`. For a ViewDefinition subject the SQL is server-generated - the user-fault cases are caught earlier as `ConstraintViolationException` - so an `AnalysisException` here skews server-side. No test covers this branch. The conservative option is to confine the 422 translation to the SQL paths, where the caller authored the SQL; if view subjects should get it too, the sync path needs the same treatment plus a test.

server/src/main/java/au/csiro/pathling/operations/sql/SqlExportExecutor.java
```diff
@@ -169,11 +171,18 @@ private List<String> runSql(
...
+      pipeline.execute(
+          Objects.requireNonNull(subject.preparedQuery()),
+          dataSource,
+          requestId,
+          result -> fileUrls.set(writeOutput(result, subject.name(), request, jobDirPath)));
+    } catch (final Exception e) {
+      // A 422 raised here survives the async round-trip: JobProvider renders a failed job through
+      // ErrorHandlingInterceptor.convertError, which passes a typed BaseServerResponseException
+      // through unaltered.
+      throw SqlOperationError.executionFailure(e);
```
johngrimes: The `ConstraintViolationException` branch above names the failing subject ("The subject '%s' is invalid: ..."), but this one reports only Spark's message with `expression = ["subject"]`. In a multi-subject export the caller cannot tell which subject failed analysis. Worth passing `subject.name()` through so the diagnostics name it.

server/src/main/java/au/csiro/pathling/operations/sql/SqlOperationError.java
```diff
@@ -97,6 +103,51 @@ public static UnprocessableEntityException unprocessable(
...
+   * without it. The result is bounded as well, since a suggestion list is drawn from the subject's
+   * columns and a wide dependency makes it long.
+   */
+  @Nonnull
+  private static String analyserMessage(@Nonnull final AnalysisException error) {
+    final String message =
+        requireNonNullElse(error.getSimpleMessage(), requireNonNullElse(error.getMessage(), ""));
+    return message.length() <= MAX_ANALYSER_MESSAGE_LENGTH
+        ? message
+        : message.substring(0, MAX_ANALYSER_MESSAGE_LENGTH) + "...";
```
johngrimes: Two nits: the nested `requireNonNullElse` evaluates `error.getMessage()` eagerly even when `getSimpleMessage()` is non-null, building the full plan string just to discard it - a short-circuiting conditional avoids that. And the truncation `substring` can split a surrogate pair; cosmetic.

server/src/test/java/au/csiro/pathling/operations/sql/SqlRunProviderIT.java
```diff
@@ -302,6 +302,67 @@ void rejectsAResourceCarryingParameterOverGet(final String parameterName) {
...
+    assertThat(body)
+        .contains("\"code\":\"invalid\"")
+        .contains("\"subject\"")
+        .contains("UNRESOLVED_COLUMN")
+        .contains("no_such_col");
+    // The analyser's suggestions name the columns the subject's own dependency declares.
+    assertThat(body).contains("family_name");
+    // Spark's getMessage would append the unresolved logical plan, naming the internal
+    // request-scoped views the dependency graph was materialised under. It is not returned.
+    assertThat(body).doesNotContain("SubqueryAlias").doesNotContain("Project [");
```
johngrimes, line 328: Worth noting that the analyser sees the rewritten SQL, in which the subject's labels have been substituted with temp view names: `line/pos` after the `FROM` clause can drift from the user's text, and a qualified suggestion can still embed an internal view name - a shape these two assertions don't cover. Not blocking; maybe worth a sentence in `sql-run.md`.

johngrimes (changes requested):
Thanks for this - the diagnosis, the write-up and the test coverage are excellent. Here are a couple of issues I found, inline below. The first two are the ones I would block on; both fixes are small, and the second simplifies `executionFailure`.

PR #2729, opened 2026-08-14
Allow $translate on Quantity

### LINKED ISSUE #2727: Terminology functions do not accept Quantity input
### Problem

The terminology functions that accept concepts — `translate`, `memberOf`, `subsumes` and
`subsumedBy` — work on `Coding` and `CodeableConcept` inputs, but fail on `Quantity`.

A UCUM-coded `Quantity` carries `system`, `code` and `unit`, which is everything needed to
identify a concept, and mapping units through a `ConceptMap` is a legitimate and useful
operation. Today an expression such as the following in a ViewDefinition column:

```
Observation.value.ofType(Quantity).translate('https://example.org/fhir/ConceptMap/units')
```

fails during analysis with:

```
Cannot convert collection of type QuantityCollection to TerminologyConcepts
```

There is no workaround available within FHIRPath itself. The engine has no function that
constructs a `Coding` from strings and no conditional, so the unit cannot be turned into
something translatable inside a view. Users are forced to project `system` and `code` as
plain columns and perform the mapping outside of Pathling, or to drop down to the library
API and use `to_coding` with the `translate` Spark function.

### Cause

The concept-accepting functions declare a `TerminologyConcepts` parameter, and
`FunctionParameterResolver` converts the input collection through `Collection.toConcepts()`.
Only two types implement that conversion: `CodingCollection` overrides it, and the base
implementation in `Collection` handles `CodeableConcept`. A `QuantityCollection` matches
neither, so the conversion returns empty and the resolver reports the error above.

The underlying UDF is indifferent to where the concept came from — it decodes rows against
the fixed `Coding` schema. The gap is purely in the FHIRPath type plumbing.

### Proposed change

Have `QuantityCollection` convert to concepts, by projecting a `Coding` structure from the
quantity's own fields — `system` to system, `code` to code, `unit` to display. A quantity
carries a single concept, so this is a set of independent concepts rather than a union.

This enables `translate`, `memberOf`, `subsumes` and `subsumedBy` on `Quantity`. The
functions that declare a `Coding` parameter directly — `display`, `property` and
`designation` — are out of scope, as they would require a conversion at the collection
level rather than at the concept level.

Quantities that carry a free-text `unit` with no `system` or `code` should be excluded from
the operation rather than raise an error. The original code is translated, not the
canonicalised form, since a `ConceptMap` is authored against what appears in the data.

fhnaumann:
Allows the terminology functions that operate upon concepts - `memberOf`, `subsumes`, `subsumedBy` and `translate` - to be invoked on a Quantity, taking the concept from the quantity's coded unit.

Fixes #2727

### DIFF
--- a/fhirpath/src/main/java/au/csiro/pathling/fhirpath/collection/QuantityCollection.java
+++ b/fhirpath/src/main/java/au/csiro/pathling/fhirpath/collection/QuantityCollection.java
@@ -17,11 +17,15 @@
 
 package au.csiro.pathling.fhirpath.collection;
 
+import static org.apache.spark.sql.functions.lit;
+import static org.apache.spark.sql.functions.struct;
+
 import au.csiro.pathling.errors.UnsupportedFhirPathFeatureError;
 import au.csiro.pathling.fhirpath.FhirPathQuantity;
 import au.csiro.pathling.fhirpath.FhirPathType;
 import au.csiro.pathling.fhirpath.Numeric;
 import au.csiro.pathling.fhirpath.StringCoercible;
+import au.csiro.pathling.fhirpath.TerminologyConcepts;
 import au.csiro.pathling.fhirpath.column.ColumnRepresentation;
 import au.csiro.pathling.fhirpath.column.DefaultRepresentation;
 import au.csiro.pathling.fhirpath.column.QuantityValue;
@@ -32,6 +36,7 @@ import au.csiro.pathling.fhirpath.definition.ElementDefinition;
 import au.csiro.pathling.fhirpath.definition.NodeDefinition;
 import au.csiro.pathling.fhirpath.definition.defaults.DefaultCompositeDefinition;
 import au.csiro.pathling.fhirpath.definition.defaults.DefaultPrimitiveDefinition;
+import au.csiro.pathling.fhirpath.encoding.CodingSchema;
 import au.csiro.pathling.fhirpath.encoding.QuantityEncoding;
 import au.csiro.pathling.sql.misc.QuantityToLiteral;
 import jakarta.annotation.Nonnull;
@@ -39,6 +44,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
 import org.apache.spark.sql.Column;
+import org.apache.spark.sql.types.DataTypes;
 import org.hl7.fhir.r4.model.Enumerations.FHIRDefinedType;
 
 /**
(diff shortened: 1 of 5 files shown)


site/docs/fhirpath/fhir-functions.md
```diff
@@ -144,3 +149,27 @@ requires a configured
...
+Observation.value.ofType(Quantity).translate('https://csiro.au/fhir/ConceptMap/units')
+```
+
+The original unit code is used, rather than the canonical form that Pathling
+derives for the purposes of comparison, as a ConceptMap or ValueSet is authored
+against the codes that appear within the data.
+
+A quantity that carries only a free-text `unit`, without a `system` and `code`,
+does not identify a concept. Such a quantity is excluded from the operation, in
+the same way as a Coding without a system and code.
```
johngrimes: "Excluded from the operation" is accurate for `translate`, which yields an empty result, but not for the boolean functions. `MemberOfUdf.doCall` filters invalid codings and then `anyMatch` over the empty stream returns `false` (`MemberOfUdf.java:85`, via `TerminologyUdfHelpers.isValidCoding`), and `SubsumesUdf` behaves the same way. So a free-text-unit quantity yields `false` for `memberOf`, `subsumes` and `subsumedBy`, and empty for `translate`.

Suggest wording to that effect, e.g. "Such a quantity yields `false` from `memberOf`, `subsumes` and `subsumedBy`, and an empty result from `translate` - the same behaviour as a Coding without a system and code."

fhirpath/src/main/java/au/csiro/pathling/fhirpath/collection/QuantityCollection.java
```diff
@@ -149,6 +155,50 @@ public static QuantityCollection fromNumeric(
...
+        .cast(CodingSchema.DATA_TYPE);
+  }
+
+  /**
+   * A quantity identifies a concept through its coded unit, so it can be used as the input to the
+   * terminology functions that operate upon concepts. Each quantity carries a single concept, and
+   * so the result is a set of independent concepts rather than a union.
+   *
+   * <p>Quantities that carry a free-text unit without a system and code yield a Coding that the
+   * terminology functions treat as invalid, and which is excluded from the operation.
```
johngrimes: Same imprecision as the site docs: for `memberOf`, `subsumes` and `subsumedBy` the invalid coding produces `false` rather than being excluded from the result. Suggest something like: "...treat as invalid; it yields `false` from the boolean functions and no output from `translate`."

fhirpath/src/test/java/au/csiro/pathling/fhirpath/dsl/TerminologyFunctionsDslTest.java
```diff
@@ -686,6 +688,138 @@ public Stream<DynamicTest> testTranslateFunction() {
...
+            "weight.translate('" + unitMap + "')",
+            "translate() returns the translation of the coded unit of another quantity")
+        .testEmpty(
+            "distance.translate('" + unitMap + "')",
+            "translate() returns empty for a quantity with an untranslatable unit")
+        .testEmpty(
+            "emptyQuantity.translate('" + unitMap + "')",
+            "translate() returns empty for an empty quantity")
+        .testEmpty(
+            "freeTextQuantity.translate('" + unitMap + "')",
```
johngrimes, line 772: `freeTextQuantity` is only exercised through `translate`. Adding `freeTextQuantity.memberOf(...)` expecting `false` would pin the boundary behaviour for the boolean functions, which return `false` rather than empty for an invalid coding, and would keep the documented behaviour honest.

johngrimes (changes requested):
Thanks, this is a nice, minimal change and the approach is exactly right: a single `toConcepts()` override with set semantics and a positionally correct projection onto the Coding schema. I ran the new test class locally at fde4e20 (129 tests, 0 failures). A couple of issues found, noted inline.

Two small non-blocking notes as well: adding "Fixes #2727" to the PR description will close the issue on merge, and the commit messages don't follow the `<type>: <description>` convention from CONTRIBUTING.md (moot if squash-merged).

PR #2754, opened 2026-09-13
Add spark_conf parameter to PathlingContext.create() (CLI --spark-conf parity)

### LINKED ISSUE #2669: Add explicit spark_conf dict parameter to PathlingContext.create() (parity with CLI's --spark-conf)
**Is your feature request related to a problem? Please describe.**

The CLI and the Python library API are asymmetric in how they let a user pass
extra Spark configuration:

- The CLI has a repeatable `--spark-conf KEY=VALUE` flag, backed by real logic
  in `pathling/cli/sparkconf.py`: keys are validated (`spark.`-prefixed),
  scalar values are coerced, `@file` secrets are resolved, and the three
  managed keys (`spark.jars.packages`, `spark.sql.extensions`,
  `spark.sql.catalog.spark_catalog`) are specially merged with Pathling's
  managed defaults so a user addition can't silently drop the Delta/
  library-runtime coordinates the library needs (with a warning if a managed
  package is overridden at a different version).
- `PathlingContext.create()` has no equivalent. The only lever is the `spark`
  kwarg — bring your own fully-built `SparkSession`. To set one Spark setting
  (e.g. `spark.driver.memory` when working with a large local terminology
  store) you either have to reach into the private
  `pathling.context._build_spark_session`, or hand-roll
  `pathling._spark_defaults.managed_spark_defaults()` yourself and risk
  diverging from what the CLI does.

Found this while tuning `spark.driver.memory` for local-mode terminology
operations against a full national SNOMED CT extension (see the memory
baseline discussion in #2665) - there was no supported, documented way to do
this from the Python API.

Separately, and unrelated to this specific gap: the docstring for the `spark`
parameter of `PathlingContext.create()` is out of date. It says a
user-supplied `SparkSession` "needs to include the Pathling library API JAR on
its classpath" and points to `pathling.etc.find_jar` for the "bundled" JAR
path - but `pathling.etc` doesn't exist in the codebase, and Pathling no
longer bundles a JAR; the classpath is populated via Maven coordinates
(`spark.jars.packages`) resolved through `managed_spark_defaults()`. This
should be corrected regardless of the outcome below.

**Describe the solution you'd like**

Add a `spark_conf: Optional[dict] = None` parameter to
`PathlingContext.create()`, applied only on the path where Pathling builds a
new `SparkSession` (a pre-existing SparkSession passed via `spark=` can't
retroactively apply most Spark settings anyway, since many - like driver
memory - are JVM-launch-time only). To avoid duplicating logic, promote the
pure, PySpark-free merge functions currently in `pathling/cli/sparkconf.py`
(`merge_spark_conf`, `_merge_packages`, `_merge_extensions`, `_merge_catalog`)
into the shared `pathling._spark_defaults` module (or a new sibling module),
and have both the CLI and `PathlingContext.create()` import from there. The
library-API version can drop the CLI-specific bits (TOML/flag string coercion,
`@file` secret resolution) and just accept a plain `dict[str, str]`.

While in there, fix the obsolete `spark` parameter docstring described above.

**Describe alternatives you've considered**

- Leave it as-is and document the "build your own SparkSession with
  `managed_spark_defaults()`" pattern instead of adding a parameter. This works
  today but is undiscoverable (no IDE-visible parameter) and doesn't get the
  managed-key merge protection the CLI has - a user could accidentally
  override `spark.jars.packages` and break the session.
- Have `spark_conf` silently merge onto an already-built `spark=` session too.
  Rejected: most interesting settings (memory, driver options) can't take
  effect after the JVM has started, so this would be misleading.

**Additional context**

Discovered during interactive testing with Claude Code while writing a
terminology-function smoke test against a local store built from a full UK
SNOMED CT RF2 release; referenced from a "Misc, other etc" comment on #2665.

developer-rpai:
## Summary

Closes #2669.

Adds a `spark_conf: Optional[Dict[str, str]]` parameter to `PathlingContext.create()`,
giving the Python library API parity with the CLI's repeatable `--spark-conf KEY=VALUE`
flag. This covers the exact gap from the issue: tuning `spark.driver.memory` for
local-mode terminology operations previously required reaching into the private
`pathling.context._build_spark_session`.

## What changed

- `lib/python/pathling/_spark_defaults.py`
  - Promoted the pure, PySpark-free merge functions from `pathling/cli/sparkconf.py`
    into the shared module, as suggested in the issue. The catalog-mismatch error is now
    `ValueError` instead of `CliError` so the library path doesn't depend on CLI machinery.
  - Added `validate_spark_conf()`: the library-API counterpart of the CLI's
    `validate_and_coerce` — enforces the `spark.` key prefix and string values.
- `lib/python/pathling/cli/sparkconf.py`
  - `merge_spark_conf` is now a thin wrapper delegating to the shared implementation,
    mapping `ValueError` back onto `CliError` (exit code 2). Existing CLI tests unchanged.
- `lib/python/pathling/context.py`
  - `create()` accepts `spark_conf`, validated and merged with the managed defaults
    (`warnings.warn` on managed-package version overrides) on the session-build path only.
    Passing `spark_conf` with an explicit `spark=` session, or when an active session
    would be reused, raises `ValueError` — most Spark settings are JVM-launch-time only.
  - Fixed the obsolete `spark` parameter docstring (`pathling.etc.find_jar` doesn't exist).

## Testing

- New pure unit tests in `lib/python/tests/test_spark_defaults.py` (no Spark startup).
- New test in `lib/python/tests/test_context_configuration.py`: `create()` rejects
  `spark_conf` combined with an explicit session.
- Existing `tests/cli/test_sparkconf.py` suite exercises the merge logic through the CLI
  wrapper unchanged.

```python
from pathling import PathlingContext

pc = PathlingContext.create(
    spark_conf={"spark.driver.memory": "8g"},
    terminology_mode="local",
    terminology_storage_path="/data/snomed-store",
)
```


lib/python/tests/test_context_configuration.py
```diff
@@ -120,3 +121,9 @@ def test_custom_configurations(spark_session):
     query_config = jpc.getQueryConfiguration()
     assert query_config.isExplainQueries()
     assert query_config.getMaxUnboundTraversalDepth() == 20
+
+
+def test_spark_conf_with_explicit_session_raises(spark_session):
```
johngrimes, line 128: This covers the rejection branch, but nothing exercises the path the feature exists for: `spark_conf` being validated, merged and handed to `_build_spark_session`. The active-session rejection branch is also untested.

`tests/cli/test_session.py` already has the pattern for this without starting Spark: monkeypatch `context_module._build_spark_session` to capture `extra_configs`. For `create()` you additionally need `SparkSession.getActiveSession` patched to return `None`, and the fake builder can raise a sentinel so `create()` stops before touching `spark._jvm`. I verified both of these pass against this branch:

```python
PathlingContext.create(spark_conf={"spark.driver.memory": "8g"})
# captured == {"spark.driver.memory": "8g"}

with pytest.warns(UserWarning, match=DELTA_COORDINATE):
    PathlingContext.create(spark_conf={PACKAGES_KEY: f"{DELTA_COORDINATE}:3.9.9"})
# captured[PACKAGES_KEY] contains the 3.9.9 coordinate
```

Those two, plus one for the active-session `ValueError`, would cover the new branches in `create()`.

lib/python/pathling/context.py
```diff
@@ -187,12 +193,19 @@ def create(
...
-        If a SparkSession is provided, it needs to include the Pathling library API JAR on its
-        classpath. You can get the path for the JAR (which is bundled with the Python package)
-        using the `pathling.etc.find_jar` method.
+        If a SparkSession is provided, it must already have the Pathling library runtime and
+        Delta Lake packages on its classpath, for example via the ``spark.jars.packages``
+        coordinates from :func:`pathling._spark_defaults.managed_spark_defaults`.
 
         :param spark: a pre-configured :class:`SparkSession` instance, use this if you need to
                control the way that the session is set up
+        :param spark_conf: extra Spark configuration entries to apply when Pathling builds a
```
johngrimes, line 202: The docs should follow this parameter. `site/docs/libraries/installation/spark.md` (Python tab, lines 50-70) still tells users to hand-build a `SparkSession` with the package coordinates, which is the workaround this parameter replaces; a `spark_conf` example there (driver memory for a local terminology store) would make it discoverable. The local-mode examples in `site/docs/libraries/terminology/local.md` (lines 387 and 476) also pass `spark` where `spark_conf` is now the simpler option.

Worth carrying over the caveat from `cli.md` lines 414-416 as well: in local mode the driver JVM is already running when the builder applies configuration, so `spark.driver.memory` set here has no effect unless the process is launched with `SPARK_DRIVER_MEMORY` (or `spark-submit`). Since driver memory is the motivating example in the PR description, users will hit this.

johngrimes (changes requested):
Thanks - the promotion of the merge logic into `_spark_defaults` is a clean lift and the CLI wrapper preserves the exit code 2 contract. Ran `ruff check`, `ruff format --check` and the Spark-free suites (`test_spark_defaults`, `cli/test_sparkconf`, `cli/test_session`, `cli/test_config`) locally: all clean, 115 passed. CI has not run on this branch yet, so the Spark-dependent test is unverified here.

Requesting changes for the missing tests on the `spark_conf` happy path in `create()` (inline comment). Docs should also be updated to point at the new parameter (inline comment).

PR #2755, opened 2026-09-13
Note trace() incompatibility with SQL aggregate functions

### LINKED ISSUE #2607: Document `trace()` incompatibility with SQL aggregate functions
## Summary

A FHIRPath expression containing `trace()` cannot be used as an argument to a SQL aggregate function (`sum`, `count`, `avg`, `min`, `max`, `collect_list`, `collect_set`, …). Pathling users who attempt this will see a Spark analyzer error with no clear guidance. This issue tracks adding documentation for the constraint and the workaround.

## Background

`trace()` produces a `Nondeterministic` Spark expression by design — this is what allows it to fire its side effects (collector entries and log output) on every row, even in contexts where Catalyst would otherwise elide duplicate evaluations.

Spark's analyzer enforces a hard rule: **`Nondeterministic` expressions may not appear inside SQL aggregate functions**. When a query of the form `df.groupBy(...).agg(sum(tracedColumn))` is planned, Spark raises:

```
AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION
```

This is a Spark constraint that applies to any `Nondeterministic` expression, not just Pathling's `trace()`. The same error occurs with `sum(coalesce(size(traceCol), 0))` — the nondeterminism propagates through any wrapping expression.

## Why Pathling cannot fix this

The `Nondeterministic` marker is intentional (see design doc in `openspec/changes/fix-trace-duplication/design.md`, section D8). Removing it would allow Catalyst to cache or elide trace evaluations across rows, breaking the trace contract. There is no way to make a traced expression both side-effecting and acceptable inside a SQL aggregate in Spark's current model.

The lambda-let fix introduced in #2594 halves the trace fire count for multi-reference `when/otherwise` patterns, but does not (and cannot) remove the aggregate incompatibility.

## User-facing workaround

If aggregation over a derived value is required, **move the `trace()` call upstream of the aggregation boundary**:

```python
# ❌ This fails — trace() inside an aggregate
result = pc.extract(
    patients,
    ["name.trace('t').given.count()"],  # will error if aggregated by Spark
)

# ✅ This works — inspect the trace on a non-aggregated projection first,
#    then aggregate the non-traced expression separately
traced = pc.extract(patients, ["name.trace('t').given"])   # materialise trace
counts = pc.extract(patients, ["name.given.count()"])       # aggregate safely
```

In general: evaluate and collect the `trace()` output first, then run aggregation on a non-traced form of the same expression.

## Proposed change

Add a note to the FHIRPath `trace()` documentation (once a dedicated page exists, or inline in the function reference) covering:

1. `trace()` produces a `Nondeterministic` expression.
2. Spark forbids `Nondeterministic` expressions inside SQL aggregate functions.
3. The error code users will see: `AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION`.
4. The workaround: move the `trace()` call upstream of the aggregation boundary.

## References

- Issue #2594 — trace() entry duplication fix (the change that surfaced this constraint)
- Design doc: `openspec/changes/fix-trace-duplication/design.md` §D8
- Spark error class: `AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION`

developer-rpai:
Closes #2607.

Adds a `trace()` entry to the FHIRPath function reference (`site/docs/fhirpath/index.md`) documenting that `trace()` produces a `Nondeterministic` Spark expression, which Spark's analyzer forbids inside SQL aggregate functions (`AGGREGATE_FUNCTION_WITH_NONDETERMINISTIC_EXPRESSION`), along with the workaround of evaluating `trace()` upstream of the aggregation boundary. Docs-only change.


johngrimes (approved):

developer-rpai:
Thanks for the approval, John! Is there anything else you'd like me to do here before it merges, or is it good to go?

johngrimes:
No, all good, thanks @developer-rpai

PR #2756, opened 2026-09-13
Surface library logging with --verbose flag

### LINKED ISSUE #2691: The CLI's --verbose flag surfaces no library logging
Running a command with `--verbose` produces none of the library's informational logging. The importer logs each stage it goes through, but an operator cannot see any of it, whether or not they ask for verbose output.

Reproducing it against the checked-in test fixture:

```bash
pathling --verbose import-snomed --default-dialect en-GB \
  terminology/src/test/resources/rf2-mini/international-20230601 /tmp/store
```

None of the following reaches the terminal, though each is logged at `INFO` by `SnomedRf2Importer`: the SNOMED edition and version that were detected, how many of each file's active rows resolved against the concept dictionary, how many concepts were written, and which dialect was chosen as the source of every concept's stored display.

The cause looks to be that `lib/python/pathling/cli/session.py` only ever lowers the log level and never raises it. Both `_build_quiet_spark` (line 129) and `_create_pathling_context` (line 152) act under `if not config.verbose`, the latter calling `setLogLevel("OFF")`. When verbose is on, neither runs, so the level is left at Spark's default of `WARN` and everything logged at `INFO` is dropped. The flag therefore buys a Python stack trace on failure, but no progress narration.

The resolution figures are worth singling out. They exist to tell an operator that a source was not self-contained - that a derived package was imported without the edition it extends, say - which is a legitimate thing to do but changes what the store contains. As things stand that report is written and then discarded before anyone can read it.

Two things would need deciding: whether `--verbose` should raise the level to `INFO` for Pathling's own loggers only, leaving Spark's own chatter at `WARN`, and whether the stage logging that an operator most needs should reach the console through the command line interface's own output rather than through the log at all.

developer-rpai:
Fixes #2691.

The CLI's `--verbose` flag surfaced no library logging: `_create_pathling_context` in `lib/python/pathling/cli/session.py` only lowered the Spark log level (to `OFF`) under `if not config.verbose` and never raised it. With `--verbose`, the level therefore stayed at Spark's default of `WARN`, and everything logged at `INFO` - e.g. the SNOMED importer's stage narration (detected edition/version, per-file resolution figures, concepts written, dialect chosen) - was silently dropped. The fix adds an `else` branch that calls `spark.sparkContext.setLogLevel("INFO")` when verbose is on.

Tests: added two mock-based unit tests in `lib/python/tests/cli/test_session.py` asserting the context log level is `INFO` with `--verbose` and `OFF` without it. Ran the CLI session test file: all 13 log-level/session tests pass (6 pre-existing `public_namespace` tests fail only in this sandbox because PySpark is stubbed; they pass in CI where the real dependency is installed).

---

## Rework — addressing review feedback

**Change of approach:** per review, the original `setLogLevel("INFO")` fix was ineffective on the Spark version the CLI depends on. The fix now targets Pathling's own log4j namespace instead of Spark's loggers: with `--verbose`, `_apply_log_level` calls `Configurator.setLevel("au.csiro.pathling", Level.INFO)` via the JVM gateway; without it, the Spark context level is set to `OFF` as before. Spark's own loggers are left at their defaults in both modes.

**Tests:** the two mock-based tests are replaced with real local Spark session tests (`lib/python/tests/cli/logging_probe.py`). Each test starts a real `local[1]` Spark session in a fresh subprocess, applies the exact `_apply_log_level` code path, and emits one INFO record from an `au.csiro.pathling.*` logger: `test_verbose_mode_surfaces_pathling_info_logging` asserts the record reaches the child's real stderr, `test_quiet_mode_hides_pathling_info_logging` asserts it stays hidden without `--verbose`. A smoke test first confirmed the diagnosis: with the root level at INFO, `setLogLevel("INFO")` surfaced nothing from the Pathling namespace, while `Configurator.setLevel("au.csiro.pathling", INFO)` surfaced the marker. `tests/cli/test_session.py`: 20/20 pass (real PySpark 4.0.4 + JDK 21).

**Decision on the open question in #2691:** for this PR the stage narration stays as log lines — `--verbose` raises INFO on Pathling's own logger namespace (`au.csiro.pathling`) only. Whether stage narration should become CLI output is a separate UX change.

**Repro output (#2691):** `import_snomed` run against the `rf2-mini/international-20230601` fixture with the fix applied — the stage narration now surfaces on stderr with `--verbose`:
```
INFO SnomedRf2Importer: Reading concepts from .../Snapshot/Terminology/sct2_Concept_Snapshot_INT_20230601.txt
INFO SnomedRf2Importer: Detected SNOMED CT edition 900000000000207008 version http://snomed.info/sct/900000000000207008/version/20230601
INFO DefaultDialect: Using language reference set 900000000000509007 ((not named in this release)) as the default dialect: the default of the SNOMED CT International edition.
INFO SnomedRf2Importer: Writing 200 concepts to the store
INFO SnomedRf2Importer: .../sct2_Description_Snapshot-en_INT_20230601.txt: 404 of 404 active rows resolved against the concept dictionary.
INFO SnomedRf2Importer: .../sct2_Relationship_Snapshot_INT_20230601.txt: 199 of 203 active rows resolved against the concept dictionary.
INFO SnomedRf2Importer: Computing the transitive closure
INFO SnomedRf2Importer: .../der2_Refset_SimpleSnapshot_INT_20230601.txt: 3 of 3 active rows resolved against the concept dictionary.
INFO SnomedRf2Importer: .../der2_cRefset_AssociationSnapshot_INT_20230601.txt: 4 of 4 active rows resolved against the concept dictionary.
INFO SnomedRf2Importer: Import complete for http://snomed.info/sct version http://snomed.info/sct/900000000000207008/version/20230601 (200 concepts)
```
The same import run without `--verbose` completes with zero narration lines — Spark's own loggers stay quiet in both modes.


johngrimes (changes requested):
Requesting changes. The fix does not take effect on the Spark version the CLI depends on, so `--verbose` still surfaces nothing. To get this merged, please:

1. Replace the `setLogLevel("INFO")` call with an explicit level on Pathling's logger namespace, e.g. `spark._jvm.org.apache.logging.log4j.core.config.Configurator.setLevel("au.csiro.pathling", Level.INFO)`. Leave Spark's own loggers at their default.
2. Replace the two fake-based tests with one that starts a real local Spark session (as the other Python tests do), emits an INFO record from a `au.csiro.pathling.*` logger under `--verbose`, and asserts it reaches stderr. A companion test that the same record does not appear without `--verbose` would also be useful.
3. Run the repro from #2691 against `terminology/src/test/resources/rf2-mini/international-20230601` and paste the resulting stage output in the PR description.
4. State in the PR description whether you think the stage narration should stay as log lines or become CLI output (the second open question in #2691). Either answer is fine for this PR; it just needs to be a decision rather than an omission.

Why the current change is a no-op: the CLI pins `pyspark>=4.0.0,<4.1.0`. In Spark 4.0, `SparkContext.setLogLevel` goes through `Utils.setLogLevelIfNeeded`, which only acts when the requested level differs from the current root level ([SparkContext.scala#L403](https://github.com/apache/spark/blob/branch-4.0/core/src/main/scala/org/apache/spark/SparkContext.scala#L403), [Utils.scala#L2339-L2343](https://github.com/apache/spark/blob/branch-4.0/core/src/main/scala/org/apache/spark/util/Utils.scala#L2339-L2343)). The root logger is already `INFO`. What hides INFO output is the `SparkShellLoggingFilter` on the console appender, installed at `WARN` because PySpark launches as an interpreter ([Logging.scala#L361-L384](https://github.com/apache/spark/blob/branch-4.0/common/utils/src/main/scala/org/apache/spark/internal/Logging.scala#L361-L384), [#L491-L513](https://github.com/apache/spark/blob/branch-4.0/common/utils/src/main/scala/org/apache/spark/internal/Logging.scala#L491-L513)). `setLogLevel("INFO")` short-circuits, so the filter stays. The issue's "left at WARN" diagnosis was inexact; sorry for the misdirection.

Reproduced on pyspark 4.0.4, inspecting JVM state through py4j:

```
root level: INFO threshold before: WARN
threshold after setLogLevel(INFO): WARN
26/09/17 16:23:34 INFO X: LIB INFO after Configurator.setLevel(au.csiro.pathling, INFO)
```

An INFO record emitted straight after `setLogLevel("INFO")` produced nothing; the same logger's INFO after `Configurator.setLevel` did, and a `DAGScheduler` INFO in the same run stayed hidden. An explicitly configured logger bypasses the shell filter (`loggerWithCustomConfig`, Logging.scala#L508), which is why item 1 works and also keeps Spark's chatter out of the way.

PR #2758, opened 2026-09-14
Key an $sql-export job on the definitions it will execute

### LINKED ISSUE #2757: $sql-export matches kick-offs to existing jobs without considering the inlined context, returning stale results
**Describe the bug**

An async `$sql-export` kick-off is matched against an already-running or finished job on
the subject's `name` and its inline SQL. The `context` parameters — the inlined
`Library` / `ViewDefinition` bodies that the subject's `relatedArtifact` chain resolves
against — are not part of that match, so the kick-off returns the earlier job's id and
that job's rows.

When the subject SQL selects from a view rather than from the data directly, the SQL is
a constant and the whole meaning of the request lives in `context`. Two kick-offs that
repeat a `name` and subject SQL while inlining different view bodies are then the same
job, and the second silently gets the first one's result: a normal `202` → `303` →
manifest, with nothing to indicate that the definitions submitted were not the ones
executed.

`$sql-run` does not behave this way — it re-resolves the inlined bodies on every request
and returns the correct result for each.

Observed on server `3.1.0-SNAPSHOT+e366202`.

This is not #2529: there the `$job` ETag was constant across distinct job ids. Here the
kick-off itself returns the *same* job id, and the ETag varies with the same key the
matching uses.

**To Reproduce**

The script below needs a base URL and two Delta tables registered under
`pathling.sqlQuery.externalTables`, each with a `patient_key` column:

```
python repro.py --base-url http://localhost:8080/fhir \
    --table-one table_one --table-one-url http://example.org/derived/table_one \
    --table-two table_two --table-two-url http://example.org/derived/table_two
```

Each step is `POST /$sql-export` with `Prefer: respond-async`, polled to `303` and
downloaded. The subject `Library` is always
`SELECT count(DISTINCT patient_key) AS n FROM crit_a`, declaring a single
`relatedArtifact` on `urn:pathling-repro:crit_a`. Only the one `context` `Library`
inlined at that url varies:

- **body A** — `SELECT patient_key FROM table_one`, declaring `table_one`
- **body B** — `SELECT patient_key FROM table_two WHERE 1 = 0`, declaring `table_two`

Body B's `WHERE 1 = 0` makes its correct answer `0` whatever the tables hold, so the
outcome does not depend on the data the server was pointed at.

```
  1. name=J  sql=S   body A                  expect n>0    2.4s  n=100  job=50a67e57-…
  2. name=J  sql=S   body B                  expect 0      0.2s  n=100  job=50a67e57-…
  3. name=J2 sql=S   body B  (truth for 2)   expect 0      2.1s  n=0    job=d0442ebe-…
  4. name=J  sql=S'  body B  (SQL + a space) expect 0      2.1s  n=0    job=128b948d-…

Reverse order, fresh name:
  5. name=K  sql=S   body B                  expect 0      2.1s  n=0    job=1e58933c-…
  6. name=K  sql=S   body A                  expect 100    0.1s  n=0    job=1e58933c-…
```

Step 2 submits a body that selects no rows and is handed step 1's job id and its 100
rows. Step 3 shows `0` is the correct answer for that body, and step 4 shows that
changing the subject SQL by a single space is enough to get a new job — which is what
identifies the match as `(name, subject SQL)` with `context` excluded. Steps 5 and 6 run
the same pair in the opposite order under a fresh name and go stale in the other
direction, so this is not specific to one table or one ordering.

**Expected behavior**

A kick-off whose inlined `context` differs from an earlier one should get its own job and
be executed against the definitions it actually submitted, rather than be matched to the
earlier job.

**A second symptom, characterised further than when I first wrote this**

I originally noted in passing that a server which has accumulated jobs this way
intermittently answers a kick-off with
`400: SQL references an unauthorised data source: LogicalRelation`. I can now say more
about when it happens, though still not why.

Four runs of the same client pipeline, sequentially, against one server and one dataset.
Each run inlines its whole closure in `context`; runs 2 and 4 are interactive, so a
researcher edited a view body partway through and the pipeline re-submitted the same
canonical `context` urls carrying different SQL.

| run | re-submitted a changed body under a `(name, subject SQL)` the server had already seen | outcome |
|---|---|---|
| 1 | no | all probes and the final export succeeded |
| 2 | yes, after 6 edits | first kick-off over the edited closure → `400 … LogicalRelation` |
| 3 | no | all probes and the final export succeeded |
| 4 | yes, after 7 edits | probes kept returning the pre-edit count; the final export → the same `400` |

Run 3 ran between the two failures and was clean, so this is not a server that degrades
and stays degraded. The two failures are exactly the two runs that re-submitted a changed
body under a name the server had already seen.

Run 4 is the informative one. Its probes carry a constant subject SQL
(`SELECT count(*) AS n … FROM (SELECT * FROM cohort) t`) under a constant name, so after
each edit they matched an existing job by the key above and returned an unchanged count.
The final export carries `SELECT * FROM cohort` — a subject SQL that name had never been
submitted with — so it was the first request in that run that could not match, and
therefore the first one forced to actually execute the edited closure. That is the one
that was rejected. The unchanged counts are also consistent with a genuine no-change in
that cohort, so I cannot offer them as proof of staleness; the point is that the client
artifacts cannot tell the two apart, which is the same silence this report is about.

My working hypothesis is that the trigger is a canonical `context` url re-submitted with
different content, rather than a repeated job name as such. I have not isolated it
server-side and could not provoke the `400` from the standalone script above, so this is
offered as characterisation rather than as a second confirmed defect.

I have made two client-side changes, in case they help narrow where to look. The first
folds a digest of the inlined closure into `subject.name`; that is step 4 of the repro
done deliberately, so it reliably avoids the stale results above. The second
content-addresses the canonical urls the `context` bodies are inlined under, so a changed
body never reuses a url — that one follows from the hypothesis and I cannot yet say
whether it removes the `400`. I will report back either way. Neither should be necessary,
which is why the report above stands as filed.

<details>
<summary><code>repro.py</code></summary>

```python
"""Standalone reproduction: `$sql-export` reuses a finished job across kick-offs
whose inlined `context` differs.

Depends only on `httpx`. Point it at a Pathling server that has two Delta tables
registered under `pathling.sqlQuery.externalTables`, each with a `patient_key`
column:

    python pathling_repro_sql_export_job_reuse.py \
        --base-url http://localhost:8080/fhir \
        --table-one chemistry \
        --table-one-url http://example.org/derived/chemistry \
        --table-two icustay_detail \
        --table-two-url http://example.org/derived/icustay_detail

Every request is read-only and aggregate-shaped.
"""

from __future__ import annotations

import argparse
import base64
import json
import time
from typing import Any

import httpx

VIEW_URL = "urn:pathling-repro:crit_a"
SUBJECT_SQL = "SELECT count(DISTINCT patient_key) AS n FROM crit_a"
LIBRARY_TYPE = {
    "coding": [
        {
            "system": "https://sql-on-fhir.org/ig/CodeSystem/LibraryTypesCodes",
            "code": "sql-query",
        }
    ]
}
VIEW_TYPE = {
    "coding": [
        {
            "system": "https://sql-on-fhir.org/ig/CodeSystem/LibraryTypesCodes",
            "code": "sql-view",
        }
    ]
}


def _library(sql: str, related: list[dict[str, str]], *, type_coding: dict[str, Any], url: str = "") -> dict[str, Any]:
    body: dict[str, Any] = {
        "resourceType": "Library",
        "status": "active",
        "type": type_coding,
        "content": [
            {
                "contentType": "application/sql",
                "data": base64.b64encode(sql.encode("utf-8")).decode("ascii"),
            }
        ],
        "relatedArtifact": related,
    }
    if url:
        body["url"] = url
        body["name"] = "crit_a"
    return body


def _kickoff_body(name: str, subject_sql: str, view_sql: str, table: str, table_url: str) -> dict[str, Any]:
    """One `$sql-export` kick-off reading `crit_a`, whose body is `view_sql`."""
    subject = _library(
        subject_sql,
        [{"type": "depends-on", "label": "crit_a", "resource": VIEW_URL}],
        type_coding=LIBRARY_TYPE,
    )
    view = _library(
        view_sql,
        [{"type": "depends-on", "label": table, "resource": table_url}],
        type_coding=VIEW_TYPE,
        url=VIEW_URL,
    )
    return {
        "resourceType": "Parameters",
        "parameter": [
            {
                "name": "subject",
                "part": [
                    {"name": "name", "valueString": name},
                    {"name": "subjectResource", "resource": subject},
                ],
            },
            {"name": "context", "resource": view},
            {"name": "_format", "valueCode": "ndjson"},
        ],
    }


def _resolve(base: str, location: str) -> str:
    return str(httpx.URL(base).join(location))


def export(
    client: httpx.Client, base_url: str, name: str, subject_sql: str, view: tuple[str, str, str]
) -> tuple[Any, str]:
    """Kick off, poll to 303, download; return the single `n` value and the job id."""
    resp = client.post(
        f"{base_url}/$sql-export",
        json=_kickoff_body(name, subject_sql, *view),
        headers={"Prefer": "respond-async", "Accept": "application/fhir+json"},
    )
    resp.raise_for_status()
    status_url = _resolve(base_url, resp.headers["Content-Location"])
    job_id = httpx.URL(status_url).params.get("id", status_url)

    while True:
        poll = client.get(status_url, headers={"Accept": "application/fhir+json"})
        if poll.status_code == 303:
            result_url = _resolve(status_url, poll.headers["Location"])
            break
        poll.raise_for_status()
        time.sleep(1)

    # `Accept: application/fhir+json` is required here: without it the manifest
    # comes back in a flattened non-FHIR shape.
    manifest = client.get(result_url, headers={"Accept": "application/fhir+json"}).json()
    locations = [
        part["valueUri"]
        for param in manifest.get("parameter", [])
        if param.get("name") == "output"
        for part in param.get("part", [])
        if part.get("name") == "location"
    ]
    ndjson = "\n".join(
        client.get(_resolve(result_url, location), headers={"Accept": "application/x-ndjson"}).text
        for location in locations
    )
    return json.loads(next(line for line in ndjson.splitlines() if line.strip()))["n"], job_id


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument("--base-url", required=True, help="Server FHIR base, e.g. http://localhost:8080/fhir")
    parser.add_argument("--table-one", required=True)
    parser.add_argument("--table-one-url", required=True)
    parser.add_argument("--table-two", required=True)
    parser.add_argument("--table-two-url", required=True)
    parser.add_argument("--name", default="pathling-repro-j1")
    args = parser.parse_args()

    base = args.base_url.rstrip("/")
    # Body B's `WHERE 1 = 0` makes the expected answer 0 whatever the data holds,
    # so the outcome does not depend on which tables the server was pointed at.
    body_a = (f"SELECT patient_key FROM {args.table_one}", args.table_one, args.table_one_url)
    body_b = (f"SELECT patient_key FROM {args.table_two} WHERE 1 = 0", args.table_two, args.table_two_url)

    with httpx.Client(timeout=httpx.Timeout(connect=10.0, read=300.0, write=300.0, pool=10.0)) as client:

        def run(tag: str, name: str, sql: str, view: tuple[str, str, str]) -> Any:
            started = time.monotonic()
            value, job_id = export(client, base, name, sql, view)
            print(f"  {tag:<58} {time.monotonic() - started:5.1f}s  n={value:<4} job={job_id}")
            return value

        print("Body A reads table_one unfiltered. Body B reads table_two under `WHERE 1 = 0`.")
        print("Both are inlined at the same context url, so only the context differs.\n")
        n = run("1. name=J  sql=S   body A                  expect n>0", args.name, SUBJECT_SQL, body_a)
        run("2. name=J  sql=S   body B                  expect 0", args.name, SUBJECT_SQL, body_b)
        run("3. name=J2 sql=S   body B  (truth for 2)   expect 0", args.name + "b", SUBJECT_SQL, body_b)
        run("4. name=J  sql=S'  body B  (SQL + a space) expect 0", args.name, SUBJECT_SQL + " ", body_b)
        print("\nReverse order, fresh name:")
        run("5. name=K  sql=S   body B                  expect 0", args.name + "r", SUBJECT_SQL, body_b)
        run(f"6. name=K  sql=S   body A                  expect {n}", args.name + "r", SUBJECT_SQL, body_a)

        if not n:
            print(f"\nNOTE: {args.table_one} returned no rows, so steps 1 and 6 cannot discriminate.")


if __name__ == "__main__":
    main()
```

</details>

johngrimes:
An async `$sql-export` kick-off named its dependencies by canonical URL alone, so a request that inlined a different `context` body at a URL it had used before matched the earlier job and was answered with that job's rows.

`SqlExportSupport.describe` rendered a subject as `name:kind:sql:bindings:topLevelKeysByLabel`, and `topLevelKeysByLabel` maps a SQL table label to a dependency's canonical key, which is `url` or `url|version` and carries no content. Two kick-offs repeating a subject over different inlined bodies therefore produced identical cache keys.

`ResolvedDependency` now declares `describeContent()`, implemented by all three node types, and the cache key carries a `dependencies` section rendering every resolved node's content ordered by canonical key. The key is built from the resolved graph rather than the submitted `context`, so an edited stored SQLView and a repointed external table separate jobs too.

The new integration test drives the reported sequence over HTTP: two kick-offs sharing a subject name and subject SQL, each inlining a different body at one canonical url, now get distinct jobs and each returns the rows its own body implies. It fails on `release/server/3.1.0` with both kick-offs sharing a job id.

The `400: SQL references an unauthorised data source: LogicalRelation` reported as a second symptom is a separate defect and is not addressed here.

Resolves #2757.


piotrszul (commented):
Reviewed the cache key change, the three `describeContent()` implementations, the unit tests, the
IT and the docs update, along with the surrounding resolution code (`SqlDependencyResolver`,
`ResolvedDependencyGraph`, `SqlExportRequestParser`, `RequestTagFactory`).

The fix addresses the reported defect, and the invariants it rests on hold:

- `ResolvedDependencyGraph.nodesByKey` is the resolver's shared memo map — `parseSubjects` threads
  one `LinkedHashMap` through every subject — so `describeDependencies` covers the transitive
  closure of the whole request, including nodes reachable only through another SQLView.
  `computeIfAbsent` cannot mask a conflict, since one canonical key resolves exactly once per
  request.
- The rendering is deterministic, which matters twice over: for legitimate deduplication, and
  because `resolveOwnJob` recomputes the key to find its own job. `TreeMap` normalises the closure
  order, `childKeysByLabel` / `topLevelKeysByLabel` are `LinkedHashMap`s ordered by reference order
  in the SQL, and `FhirView` is `@Data` all the way down with `ConstantDeclaration.value` a HAPI
  `PrimitiveType` whose `toString()` is value-based — so no identity hash leaks into the key.
- The identity-direction tests (`anIdenticalRequestProducesTheSameKey`,
  `anIdenticalViewDefinitionDependencyProducesTheSameKey`,
  `theSameClosureReachedInADifferentOrderProducesTheSameKey`) guard the opposite failure mode,
  where every kick-off would fork a new job.

One follow-up, not a blocker for this PR: the new `dependencies=[...]` section is built by
concatenating client-controlled values with unescaped delimiters
(`entry.getKey() + '=' + entry.getValue()`, joined with `,`), so two different resolved graphs can
still render to the same key. A payload placed after a `--` line comment survives the parse that
`validateStatically` runs at kick-off, and both the canonical keys and `childKeysByLabel` are
forgeable as well as the SQL. Impact is low — the ownership check in `runExport` blocks cross-user
reuse, and the pre-existing `subjects=[...]` section has the same property, so this is not a
regression introduced here.

Raised as #2768 with a worked example and a proposed fix (percent-encode each atom, in both
sections).

PR #2760, opened 2026-09-14
Accept a subquery over a nested SQLView's own dependency

### LINKED ISSUE #2759: A subquery inside a nested SQLView body is rejected as an unauthorised data source
A `SQLView` whose SQL uses a subquery over one of its own declared dependencies is rejected with `400 SQL references an unauthorised data source: LogicalRelation` when it is reached from a parent query, even though every relation involved is a legitimate request-scoped temp view.

`SqlValidator.walkExpressionAnalyzed` restarts the trusted-alias flag at `false` for a subquery's plan (`server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlValidator.java:624`), so trust is not inherited into a subquery from the enclosing view. `SqlQueryExecutor` then passes only the top-level labels' temp view names into `validateAnalyzed` (`server/src/main/java/au/csiro/pathling/operations/sqlquery/SqlQueryExecutor.java:132-133`). A subquery sitting inside a nested view's expanded plan references that nested view's own child temp view, whose name is not in the set, so `isTrustedAlias` does not recognise it and the `LogicalRelation` beneath it is rejected.

The node passes its own check when it is materialised, because `materialiseNode` supplies that node's `childViewNames` (`SqlQueryExecutor.java:169`). The rejection happens only at the enclosing query's check.

**To Reproduce**

Against a graph of an external table leaf `t`, a SQLView `v` whose SQL is `SELECT * FROM t WHERE id >= (SELECT min(id) FROM t)` declaring `t`, and a top-level query `SELECT * FROM v` declaring `v`, `SqlQueryExecutor.execute` throws:

```
InvalidRequestException: SQL references an unauthorised data source: LogicalRelation
```

The same body without the subquery succeeds.

**Expected behavior**

A subquery over a declared dependency is ordinary SQL and should be accepted. The check appears to need the full set of registered temp view names rather than only the top-level labels', or trust should be inherited into subquery plans; which of the two preserves the guarantee the check exists for needs deciding deliberately, since it widens what the analysed-plan gate accepts.

Found while investigating the second symptom described in #2757. It plausibly explains the report there: the probes in that run carried a constant subject SQL and were matched to an existing job, so the edited closure was never executed, and the final export was the first request forced to execute it.

johngrimes:
The analysed-plan trust flag restarted at `false` for a subquery's plan, so trust was not inherited from the enclosing view. A `SQLView` body that used a subquery over one of its declared dependencies was rejected as an unauthorised data source once a parent query reached it, even though the node had passed the same check when it was materialised.

`walkExpressionAnalyzed` now carries the enclosing plan node's trust into any subquery plan it encounters, so a subquery inherits trust exactly as the node's plan children do. A subquery written in the requester's own top-level SQL still starts untrusted, because the node carrying it sits above every trusted alias, so the gate does not widen for requester-written SQL.

The IT fixture's Patient data is now read back from a Parquet file, so a ViewDefinition leaf's plan carries a `LogicalRelation` as it does against the Delta warehouse. From an in-memory dataset it carries a `LocalRelation`, which the trust gate does not police, and the ITs could not observe the defect at all.

Resolves #2759.


piotrszul (approved):
## Summary

A targeted fix to the analysed-plan trust gate in `SqlValidator`: the trust flag is now carried into subquery expression plans instead of being reset. The change is small, correctly reasoned, and comes with a positive test, a negative control, and an IT fixture change that makes the defect observable.

## Suggestions (non-blocking)

1. **`SqlViewTestConfiguration.replacePatientsWithFileBacked`** — the Parquet fixture is written to a hard-coded relative path under `target/`. This works under Maven and is safe today because this configuration is imported by only one IT class, but it is CWD-dependent and would silently become a race if a second IT class ever imported it with failsafe's `forkCount=2`. Consider deriving the path from `java.io.tmpdir` or the `project.build.directory` system property, and noting the single-importer assumption in the Javadoc.

2. **Pre-existing, not introduced here** — the analysed gate trusts anything beneath a `SubqueryAlias` whose name matches a registered temp view, and temp view names are derived from a caller-supplied `X-Request-ID`. A requester could alias their own derived table to that name. Today this is unexploitable because the parse-time strict walk rejects any undeclared relation first, and this PR does not widen that. Worth a separate issue if the analysed check is ever meant to stand alone.

## What looks good

- **The right one of the two fixes proposed in #2759.** Trust is positional, so a subquery hanging off a node beneath a trusted alias is beneath that alias too. Passing the full set of temp view names to the top-level check would have let top-level SQL name nested views directly.
- **No widening for requester-written SQL.** `validateAnalyzed` still enters with `inTrustedAlias=false`, and a subquery in top-level SQL hangs off a Filter/Project that sits above every trusted alias, so it still starts untrusted. `rejectsSubqueryOverUnregisteredRelationOutsideTrustedAlias` pins this down.
- **Defence in depth is preserved.** The strict parse-time walk still descends into subqueries with the declared label set, and the executor still validates each SQLView node against its own children at materialisation before the top-level check.
- **Label rewriting already handles subqueries** (`SqlLabelRewriter`), so the fixture view's `ap` reference inside the scalar subquery is rewritten correctly.
- **The IT fixture change is the most valuable part.** In-memory datasets produce a `LocalRelation`, which the gate does not police, so the ITs previously could not observe this class of bug at all. Round-tripping Patient through Parquet gives them a `LogicalRelation` as in production.

## Verification

| Check | Result |
|---|---|
| `SqlValidatorTest` (local) | 111 passed, both new cases included |
| `SqlViewRunProviderIT` (local) | 18 passed, new subquery case included |
| CI: Build and test server | pass |
| CI: SonarCloud | pass |

Issue #2761, opened 2026-09-17
Parquet on FHIR M1 — The layout


piotrszul:
Part of #2367. The first of seven milestones in the Parquet on FHIR delivery.

The contract is the specification bundle at `specs/001-parquet-on-fhir/`, which
lives on the `issue/2367` baseline branch. This milestone covers **phases 1–6,
69 tasks**, all of them done; the pull request is #2767.

### Delivers

FHIR JSON to the new layout and back through `io`, losslessly except the
exceptions FR-016 lists. No annotations, no primitive ids or extensions, and the
pruned schema mode only. One definition traversal, the canonical structure and
the one shared merge. Both risk gates resolved.

The definition abstraction and the traversal belong here rather than to the
engine, because the ingest transform writes the layout the definitions describe
and FR-008/FR-012 forbid taking types or cardinality from the data.

### Narrowed since this issue was raised

**Decision 68** narrowed the milestone to the value layer. Decimals are stored
numerically rather than lexically, and standard Spark JSON read and write are
used with a double as the serde intermediate; `base64Binary` becomes binary,
which the Parquet on FHIR specification requires and the existing encoder
already does. Content outside the definition set is ignored with a warning
rather than being switchable between ignoring and failing, so the strictness
switch goes. Value validation, dense-bounds reporting and the large synthetic
round trip are deferred to later phases.

It also collapsed schema derivation into the ingest transform: `SchemaBuilder`
and `SchemaPruner` are deleted, one traversal of the definitions builds the
columns, and the stored schema is the result of that transform rather than a
value built alongside it and kept in step.

**Decision 69** moved the dense schema mode out of this milestone and into M6,
after the annotations and the primitive metadata group. With the derivation
deleted, the mode's absence leaves no second derivation behind: what ships here
is one traversal with one strategy, and the second strategy is additive.
FR-009, FR-010 and FR-044 are therefore delivered in M6 rather than here.

[...]

Issue #2766, opened 2026-09-17
Parquet on FHIR M6 — Completion


piotrszul:
Part of #2367. The last of six milestones in the Parquet on FHIR delivery.

The contract is the specification bundle at `specs/001-parquet-on-fhir/`, which
lives on the `issue/2367` baseline branch. This milestone covers **phase 17,
15 tasks**.

### Delivers

The previous-layout reader removed, FR-053 tightened, the superseded design
retired, and the follow-ups raised along the way closed out.

### User-visible change

None, **contingent on T049a**: if the source boundary routes earlier-layout data
rather than refusing it, the reader T100e removes is a product feature and its
removal is breaking.

### Depends on

#2765.

### Delivery

The pull request targets `issue/2367`, not `main`, so that the milestones stack
and the baseline branch carries only the specification. Each milestone's pull
request is subject to the adversarial review in the `ssh:build` process, scoped
to that milestone's phases.

Issue #2367, opened 2025-06-04
Implement Parquet on FHIR


johngrimes:
It would be great if we could replace our current encoding scheme with the one described here: https://github.com/aehrc/parquet-on-fhir

This would provide a number of benefits, including true lossless round-trip of FHIR data, improved handling of decimals, dates and quantities, and the ability to use smaller and more performant schemas.

johngrimes:
Let's take a look at the variant data type as part of this work: https://medium.com/@krthiak/variant-data-type-in-spark-day-55-of-100-days-of-data-engineering-ai-and-azure-challenge-6a935af86778

This also relates to #2356.

johngrimes:
Here is the branch that I used to partially implement some of the concepts: https://github.com/aehrc/pathling/tree/spike/parquet-on-fhir

johngrimes:
Is there a gap in the spec regarding how to represent multiple values within an annotated field?

PR #2767, opened 2026-09-17
Parquet on FHIR M1: The layout

piotrszul:
Resolves #2761, the first of seven milestones for #2367 (Parquet on FHIR).

This targets `issue/2367`, which holds the specification bundle at `specs/001-parquet-on-fhir/` and nothing else. The milestones stack on it; decision 65 records why. Merging here does not close #2761 automatically, because the base is not the default branch.

> **Status: all 68 of M1's tasks are done.** The milestone was rescoped by decisions 68 and 69 after a first implementation was complete, and the 21 tasks that rescoping reopened are closed. Decision 70 then split `io`'s public API, adding T080a. The code now implements the specification in this branch. The adversarial review below was of the first implementation; the rescoped code has not yet been through it.

## What M1 delivers

FHIR JSON goes into the Parquet on FHIR layout and back out again, losslessly except for the exceptions FR-016 lists. No annotations, no primitive ids or extensions, and the pruned schema mode only. Phases 1–6.

- **`fhir-schema`** (new module): the definition abstraction, moved here from `fhirpath` so that `io` can reach it without depending on the engine, plus the canonical structure and the storage type of each primitive. The default definitions keep the field name their accessor is derived from, and implement the interface's cardinality method explicitly, so that the relocation stays the engine's only change.
- **`utilities`**: the one structure merge and its interface, over `spark-sql-api` types only.
- **`io`** (new module):
  - `ResourceTransformer` works on datasets in both directions and reads or writes no text. `toLayout` takes FHIR JSON already parsed into a dataset and builds every stored column in one walk over the definitions; the input schema decides presence and nothing else, and the stored schema is whatever the columns produce. `toJsonShape` returns the layout as a dataset of plain JSON types for the JSON writer.
  - `FhirJsonReader` reads files or a `Dataset<String>` and applies `toLayout`; text that is not JSON fails the read. `FhirJsonWriter` returns a `Dataset<String>` or writes files directly, leaving null fields out of both.
  - `PrimitiveConverters` has a named converter for each of the 21 primitives, stating the column types it accepts, what it stores, and what egress writes.
  - `StrictnessCheck` reports what the layout does not store, which is ignored with a warning and never fails the read.
- **`fhirpath`**: definition classes relocated to `fhir-schema`, and a dependency on it in the POM. The source is otherwise identical to `main`, and the module's diff is imports.
- **`benchmark`**: separate encode and decode states, so M4's measurement has a baseline to compare against.
- **Build**: a checkstyle rule keeps Catalyst imports out of `io`, `fhir-schema` bans the engine as a dependency, and `library-runtime` declares both new modules explicitly so the Python and R libraries ship them.

## What the two late decisions changed

**Decision 68 narrowed M1 to the value layer.**
- **Decimals** are read and written through a double. They are stored as text, as the Parquet on FHIR specification requires, and are numerically equal to the source rather than textually identical to it. `evidence/scripts/double_serde.py` measures what each form stores and writes.
- **`base64Binary`** is stored as binary. That is the Parquet on FHIR mapping, and the existing encoder already does it.
- **Ignored content**: content outside the definition set, or whose shape or JSON encoding contradicts the definitions, is ignored with a warning. There is no longer a switch to make it fail.
- **Deferred**, each with a named follow-up: value-range validation, dense-mode bounds reporting and the large synthetic round trip.

It also **collapsed schema derivation into the transform**. `SchemaBuilder`, `SchemaPruner`, `SchemaConfiguration`, `BoundsCheck`, `DecimalTransform` and `PrimitiveValues` are deleted.

**Decision 69 moved the dense schema mode to M6**, a milestone of its own after the annotations and the primitive metadata group (#2771). With the derivation gone, the mode's absence leaves no second derivation behind. The three bounds options stay in the public API and bound nothing between the flip and M6; the window is documented in `contracts/library-api.md`. Completion renumbers to M7 and keeps #2766.

**Decision 70 split `io`'s public API** into the transformer above and a reader and writer over JSON text. `PathlingContext.encode` takes a `Dataset<String>` and `decode` returns one, and `io` previously read only paths, so M4 had no entry point to call. With the split, encode, decode and the NDJSON sink are one call each, and M3's bundle and XML ingest reach the same `toLayout` without changing it.
- **Input contract**: the types Spark's JSON reader infers, plus lossless widenings a hand-built dataset would carry. The integer types and `decimal` accept Byte, Short, Int and Long columns. `decimal` refuses Float, which has already rounded the value, and the temporal types refuse Spark Date and Timestamp, which cannot hold a partial date or an offset. Anything else is still reported and nulled.
- **One resource type per input** is documented and not checked. Today's `encode` silently drops documents of other types, so T070 is reworded to keep that selection in `PathlingContext` rather than in `io`.
- `toJsonShape`'s output holds decimals as doubles, so it is for serialising and not for storing, and its javadoc says so.

## How the rescoped work landed

The branch was torn down first, then rebuilt. Classes and tests asserting withdrawn behaviour were deleted in one commit, before any replacement was written, so that no test carries a withdrawn behaviour forward by inertia.

- **The traversal** plans each column once per file and applies the plan to every row. An element the source does not carry does not appear. A structure appears only where something beneath it is stored (FR-011). An element supplied in the wrong shape is stored as a null of the shape the definitions declare. That type is built from columns, not declared as a `StructType`. A structure supplied as a plain value, which leaves nothing to build a type from, is dropped and reported.
- **The converter table** checks each column, not each value. A column whose inferred type the converter does not accept is nulled for the whole file and reported as an `encodingMismatch`. So a fraction among integers voids the element rather than being truncated, and `"yes"` for a boolean becomes null rather than `true`.
- **One addition to decision 68**, recorded as an addendum: `decimal` also accepts `DecimalType`. Spark infers `decimal(20,0)` for a file of whole-number values where one exceeds the range of a long. Rejecting it dropped conformant content that FR-016 did not name as an exception.
- **One deviation from the task text**, recorded against T060a: base64 ingest uses `try_to_binary(value, 'base64')` rather than `unbase64`. Malformed content then becomes null, as every other converter's `try_cast` does, instead of raising. A test pins that whitespace between groups still decodes.

## What users see

Nothing. `PathlingContext.encode`, `decode` and `NdjsonSink` still write and read the previous layout, and the new path is reachable only through `io`'s own entry points and the tests. `encoders`, `terminology`, `library-api`, `lib/` and `server/` are untouched. The public API switches in M4.

## Evidence

- **`io` after decision 70**, at `ff986dbed8`: 330 tests, 0 failures, 1 skipped. That is 47 more than before the split: `StructFixtureTransformTest` tests the transform on datasets built with explicit schemas, including types JSON reading never produces (13 tests); `FhirJsonReaderTest` and `FhirJsonWriterTest` check that a file and a dataset of the same text read identically, and that the written files hold exactly the documents returned, decimals included (5); and `PrimitiveConverterTest` gains 29 accepted and refused column types. The full build has not been re-run since the split; nothing outside `io` references the classes it moved or deleted.
- **Full build** before decision 70: `mvn clean install -pl library-runtime -am` at `3dd1b0fd28`: BUILD SUCCESS in about 14 minutes, 8,390 tests, 0 failures, 0 errors.

  | Module | Tests | Skipped |
  |---|---|---|
  | `utilities` | 35 | 0 |
  | `fhir-schema` | 38 | 0 |
  | `encoders` | 536 | 0 |
  | `io` | 283 | 1 |
  | `terminology` | 522 | 0 |
  | `fhirpath` | 6,808 | 924 |
  | `library-api` | 168 | 0 |

  The 924 `fhirpath` skips are the conformance exclusion baseline, the same count as before decision 68, and no exclusion was obsoleted. The `io` skip is the disabled Synthea round trip. The last full build before the rescoping, at `0141a8b19f`, ran 8,514 tests. Among the changes since, the specification-example round trip dropped from 281 tests to 168 when it moved to the subset below.
- **Round trip over the R4 specification examples**:
  - Detection runs over every type the vendored corpus carries.
  - The round trip runs over a structural subset: the 25 of 139 round-trippable types that between them cover all 79 corners `MANIFEST.tsv` names. The subset is chosen at runtime, and a test checks the coverage.
  - The class runs 168 tests in about 25 seconds, down from 281 in about 100. Restoring full coverage is a one-line change to the selection.
- **FR-016's exceptions are counted, not applied silently.** The harness reports how often each applied: primitive metadata, ignored content, base64 whitespace, and numbers that matched only numerically. Each test states the counts it expects. Undescribed content shortening an array, a column re-typed by one sibling value, and base64 whitespace each have a test of their own.
- **Round trip through Parquet on disk**: the richest specification examples and two Synthea types, with the schema read back checked against the schema written. This is the only test that exercises the Parquet layer. The large Synthea round trip is deferred by decision 68 (T077), and its class is disabled rather than deleted.

## Adversarial review

The first implementation went through the `ssh:build` review, scoped to M1's phases. Findings about later milestones' requirements were out of scope, but anything M1 does that blocks a later requirement was in scope.

- **Round 1: `NEEDS_WORK`.** A malformed primitive value crashed the transform with a raw Spark exception. Probing found worse:
  - With `spark.sql.ansi.enabled` off, which is the caller's setting and not ours, bad values were stored as null and `1.5` as the integer `1`.
  - With it on, `"yes"` was stored as boolean `true`.
  - A malformed decimal leaked internal marks on egress.

  It was fixed by checking each value against the lexical form the R4 definitions declare for its type, independently of ANSI mode.
- **Round 2: `SATISFIED`**, no required changes. The reviewer attacked the fix with its own inputs through real Parquet files.

**Decision 68 replaced that fix, and the replacement is what a reviewer should now judge.**
- The per-column check covers the cases the regexes were written for: `1.5` against an integer, `"yes"` against a boolean, and a number supplied for text all have tests.
- It is independent of ANSI mode, because every cast is a `try_cast` and the acceptance decision is made before any cast. The marks are gone with `DecimalTransform`.
- What it does **not** cover is value-range validation: `positiveInt` of `0` and `unsignedInt` of `-1` now pass. That loss is recorded on decision 57 and carried by T134g.
- Granularity moves from per value to per column, so one `1.5` among integers voids that element for every resource in the file. That is accepted, because casting per value turns `1.5` into `1`, a plausible wrong answer rather than an absence.
- Decision 57's "a value contradicting its declared type fails regardless of the switch" is overturned outright, as decision 57 itself asked to be done explicitly.

**Decisions 67 to 70, and all of the code written under them, post-date the review.** Two steps are left before this is ready: re-running the full build after the split, and re-running the review over the rescoped phases.

## Worth a reviewer's attention

- **Decision 67** withdraws FR-015 mid-milestone. The definition abstraction was to report a type as a value of its own, so that a code outside the R4 enumeration would be representable, and the engine had been converted to that type across some fifty files. Nothing in the programme needs it: a walk over every element reachable from every R4 resource finds no type the enumeration lacks, and `FhirTypeCoverageTest` now keeps it that way. What it defers is that a second provider will have to introduce a version-neutral type then, against three modules rather than one.
- **Decision 68's honest cost**, stated in the decision itself: M1's exit criterion moves from "lossless, or loud about what was lost" to "lossless for the resources we test". Detection is kept, and reported as a warning. But the corpus that would have exercised it is reduced, and the result lands on the M4 flag day, where this specification names silent loss as the worst failure mode in the programme.
- **The one exception to "no behaviour change" in `fhirpath`** is in the definitions it consumes, not in the module. A choice now expands in the order FHIR declares its types. And a choice that admits a reference carries a single reference variant, rather than one name per target resource that no instance can populate. The full suite was green and no conformance exclusion was obsoleted, but no test asserts that the engine's choice behaviour is unchanged.
- **Recorded but not fixed**, each with its reason:
  - Decision 63: an array drops null elements, which T061 must revisit.
  - Decision 64 and T134e: an inline resource, and an empty array, are dropped correctly but reported with the wrong diagnosis.
  - Decision 66: ignoring content can shorten an array. With the switch withdrawn, that is now an unconditional round-trip exception, and the harness asserts it.
- `io/src/test/resources/data/fhir-spec/R4/` is 47% of the changed files. It is vendored CC0 data, with provenance and selection documented in its `README.md`.


io/src/main/java/au/csiro/pathling/io/transform/StrictnessCheck.java
```diff
@@ -0,0 +1,160 @@
...
+      findings.add(
+          NonConformantContent.shapeMismatch(
+              path,
+              repeating
+                  ? "the definitions declare a repeating element and the source carries a single"
+                      + " value"
+                  : "the definitions declare a singular element and the source carries an array"));
+      return;
+    }
+    final DataType value = elementTypeOf(observed);
```
piotrszul, line 140: **Arrays nested inside arrays are never detected.**

The cardinality check above compares only the outermost array level against `isRepeating()`, and `elementTypeOf(observed)` then strips every array level before the type is inspected. `BoundsCheck` does the same, as does `shapeMatches` in `ResourceTransformer`.

With a source like `"name":[{"given":[["Jim"]]}]`, no finding is reported. `PrimitiveValues.storedValue` then applies `try_cast` to `array<string>` with a string target, which Spark allows, so `given` is stored as the text `"[Jim]"` with no warning or error, even with `failOnNonConformantContent` on.

For integer or boolean elements (e.g. `[[1]]`), `rlike` runs on an array and the read fails with a Spark analysis error instead of a finding. For `[[{...}]]`, the struct goes through `structure()` and fields are read from the wrong level.

Suggest reporting a shape mismatch when the element type of an array is itself an `ArrayType`, rather than unwrapping all levels.

io/src/main/java/au/csiro/pathling/io/transform/PrimitiveValues.java
```diff
@@ -0,0 +1,111 @@
...
+            ? DecimalTransform.storedValue(source, target)
+            : source.try_cast(target);
+    return Optional.ofNullable(LEXICAL_FORMS.get(type.toCode()))
+        .map(
+            form ->
+                functions
+                    .when(
+                        source
+                            .isNull()
+                            .or(source.rlike("^(" + form + ")$").and(stored.isNotNull())),
```
piotrszul, line 98: **The JSON type of a primitive value is ignored.**

Because the reader uses `primitivesAsString`, a JSON string and a JSON number/boolean are indistinguishable by the time this runs, and the lexical check only tests the text. So:

- `"active":"true"`, `"multipleBirthInteger":"5"` and `"valueDecimal":"1.50"` are accepted as boolean, integer and decimal.
- `"family":123` is accepted as the string `"123"`.

`ResourceSerialiser` writes each value by its declared type, so the output JSON differs from the input with no finding reported. That seems to be the kind of value 9113273b46 ("Refuse a value that contradicts its type") set out to refuse.

Related: the pattern is anchored as `^(...)$`, and in Java regex `$` also matches before a trailing line terminator. `"5\n"` passes as an integer (where `try_cast` trims it to 5). For a decimal the stored text keeps the newline, the serialiser's pattern then fails to match, and the value is written as a quoted string still containing the `\u0001` markers. Anchoring with `\z` instead of `$` would close that gap.

piotrszul (commented):
## Code review (at 789d13b347)

Scope: the new main code in `io` (`ResourceTransformer`, `StrictnessCheck`, the primitive converters, JSON reader and writer), `fhir-schema` (`DefinitionCanonicalStructure`, `LayoutEntry`, `LayoutFields`, `MetadataGroupStructure`, `StructureMerge`), the benchmark states, and the POM and CI wiring. I did not review the tests or the `fhirpath` package-rename churn. These findings come from reading the code; I did not build or run anything.

### 1. Nested arrays are neither rejected nor reported (medium)

[`ResourceTransformer.java:453`](https://github.com/aehrc/pathling/blob/789d13b347beb6895a0b5f54bb4952b4fa32943a/io/src/main/java/au/csiro/pathling/io/transform/ResourceTransformer.java#L453), [`ResourceTransformer.java:428`](https://github.com/aehrc/pathling/blob/789d13b347beb6895a0b5f54bb4952b4fa32943a/io/src/main/java/au/csiro/pathling/io/transform/ResourceTransformer.java#L428), [`StrictnessCheck.java:222`](https://github.com/aehrc/pathling/blob/789d13b347beb6895a0b5f54bb4952b4fa32943a/io/src/main/java/au/csiro/pathling/io/transform/StrictnessCheck.java#L222)

`elementTypeOf` unwraps arrays recursively, but `shapeMatches` and `StrictnessCheck.checkElement` only check whether the *outer* type is an array. So a repeating element that arrives as an array of arrays, such as `"given": [["a","b"]]` (inferred as `array<array<string>>`), passes both checks as conformant. `value()` then applies the per-value plan to each inner array through `functions.transform`, and no finding is raised.

- **Text primitive:** `try_cast(array<string> → string)` succeeds, so `given` is silently stored as `["[a, b]"]`. That is exactly the silently wrong value the class says it never produces.
- **Integer or boolean:** `try_cast(array<bigint> → int)` is an analysis-time type mismatch, so the whole read throws instead of the content being reported and ignored.
- **Repeating structure**, such as `"name": [[{...}]]`: `getField` on the `array<struct>` gives array-valued fields, which then fail in one of the two ways above.

Suggested fix: treat an array whose element type is itself an array as a shape mismatch in both places.

### 2. Source keys with annotation names are dropped without a finding (low)

[`StrictnessCheck.java:96`](https://github.com/aehrc/pathling/blob/789d13b347beb6895a0b5f54bb4952b4fa32943a/io/src/main/java/au/csiro/pathling/io/transform/StrictnessCheck.java#L96), [`ResourceTransformer.java:209`](https://github.com/aehrc/pathling/blob/789d13b347beb6895a0b5f54bb4952b4fa32943a/io/src/main/java/au/csiro/pathling/io/transform/ResourceTransformer.java#L209)

This concerns a source key (a JSON property, or a column of a source DataFrame) that has the same name as one of the layout's own annotation fields, such as `__birthDate_start`, `__valueQuantity_numeric` or `__x_canonical`. It does not concern FHIR's `_x` metadata groups, which are handled correctly.

```json
{ "resourceType": "Patient", "birthDate": "1970", "__birthDate_start": "whatever" }
```

1. `StrictnessCheck.check` finds `__birthDate_start` in the canonical structure, because it is an annotation entry. So the key is not reported as undescribed content. It is not a metadata group, so it goes to `checkElement`. An annotation entry carries the definition of the element it annotates, so the key is checked as if it were `birthDate`, passes, and nothing is reported.
2. `ResourceTransformer.field` returns empty for any entry where `!entry.isElement()`, so the value never reaches the output.

The content is lost and `findings()` doesn't mention it, which breaks the promise that ignored content is always detectable. An unknown key such as `"foo"` would have been reported. By the same mechanism, a structure-shaped `__valueQuantity_canonical` is walked against the Quantity definition, so any finding it produces refers to the wrong thing.

Real FHIR JSON will almost never contain such keys. The realistic routes are feeding layout-shaped data (for example the Parquet output) back through `toLayout`, or a hand-built DataFrame source that uses those names.

Suggested fix: in `StrictnessCheck.check`, report a key that matches an annotation entry as undescribed content (or a dedicated reserved-name finding) before it reaches `checkElement`.

io/src/main/java/au/csiro/pathling/io/transform/PrimitiveValues.java
piotrszul: `PrimitiveValues` is gone. Decision 68 deleted it, and the reader no longer uses `primitivesAsString`: Spark infers each column's type from the JSON, and each converter in `PrimitiveConverters` accepts only the column types it names. That settles most of this. Probed at `02c40fc562`:

| Source | Result |
|---|---|
| `"active":"true"` | ignored, reported: "declares a boolean … carries a string column" |
| `"multipleBirthInteger":"5"` | ignored, reported |
| `"gender":123`, alone in its file | ignored, reported: "carries a bigint column" |
| `"active":"true"` beside `"active":false` | column inferred as string, both ignored, reported |
| **`"gender":123` beside `"gender":"male"`** | **stored and written as `"123"`, no finding** |
| **`"birthDate":2000` beside a string date** | **written as `"2000"`, no finding** |

The trailing-newline point no longer applies: there is no lexical pattern left to anchor.

What remains is the last two rows. Spark types a column once per file, so a number beside a string for the same element widens the column to string, every string-like converter accepts it, and the source's JSON type is gone before the check runs. It is neither ignored nor reported, which FR-018 requires. It is recorded as **T124j** in M5, with the reported case and the conformant string beside it to be pinned in `StrictnessTest`. T134g's per-value validation is related but doesn't cover it, since the value is kept rather than nulled.

PR #2769, opened 2026-09-18
Make the $sql-export cache key unambiguous

### LINKED ISSUE #2768: Forgeable delimiters in the $sql-export cache key
## Problem

`SqlExportSupport.computeCacheKeyComponent` builds the `$sql-export` async job cache key by
concatenating client-controlled strings with unescaped delimiters. Two structurally different
dependency graphs can therefore render to the same key, and the second kick-off is deduplicated
onto the first job and answered with its rows.

The `dependencies=[...]` section added in #2758 renders each node as `key=content`, joined with
`,`:

```java
contentByKey.entrySet().stream()
    .map(entry -> entry.getKey() + '=' + entry.getValue())
    .collect(Collectors.joining(","));
```

and `ResolvedSqlView.describeContent()` is `"sql-view:" + sql + ':' + childKeysByLabel`. Neither
`sql` nor the canonical keys are escaped, so the grammar can be reproduced from inside a single
value.

This is not a regression introduced by #2758 — the pre-existing `subjects=[...]` section, rendered
by `describe(SubjectInput)`, concatenates `name:kind:sql:bindings:topLevelKeysByLabel` the same
way.

## Worked example

Both requests are accepted: the payload sits after a `--` line comment, so Spark's parser (run at
kick-off by `validateStatically`) sees valid SQL referencing only declared labels.

**Request A** — subject plus three `context` SQLViews:

| Artefact | SQL | `relatedArtifact` |
|---|---|---|
| subject (`sql-query`) | `SELECT * FROM a` | `a` -> `.../v1` |
| context `.../v1` | `SELECT * FROM lbl --` | `lbl` -> `.../v2` |
| context `.../v2` | `SELECT n * 10 AS n FROM lbl` | `lbl` -> `.../v3` |
| context `.../v3` | `SELECT 1 AS n` | none |

Result: `n = 10`.

**Request B** — identical subject, two `context` SQLViews:

| Artefact | SQL | `relatedArtifact` |
|---|---|---|
| subject (`sql-query`) | `SELECT * FROM a` | `a` -> `.../v1` |
| context `.../v1` | `SELECT * FROM lbl --:{lbl=https://example.org/Library/v2},https://example.org/Library/v2=sql-view:SELECT n * 10 AS n FROM lbl` | `lbl` -> `.../v3` |
| context `.../v3` | `SELECT 1 AS n` | none |

Result: `n = 1`.

Both render the same `dependencies` section, because B's `v1` declares the same children as A's
`v2`, so B's real trailing `:{lbl=.../v3}` lands exactly where A's `v2` entry ends:

```
.../v1=sql-view:SELECT * FROM lbl --:{lbl=.../v2},.../v2=sql-view:SELECT n * 10 AS n FROM lbl:{lbl=.../v3},.../v3=sql-view:SELECT 1 AS n:{}
```

The `subjects=[...]` section matches too, so the whole `operationCacheKey` matches. B is handed A's
job id, B's SQL never executes, and B's client downloads A's output files.

## Impact

Low, and bounded:

- **Not reachable across users.** `SqlExportSupport.runExport` compares `ownJob.getOwnerId()` to
  the current principal and throws `AccessDeniedError`, so a colliding request from another
  identity gets a 403 rather than another user's rows.
- **Forging buys nothing in the cross-user denial case.** Anyone able to forge a collision with a
  victim's query could submit that query verbatim and pre-own the job — the key is designed to
  collide on equivalent requests.
- **Never happens by accident.** It requires `--` followed by an exact reconstruction of the
  renderer's grammar.
- **Requires unchanged database state.** `RequestTag.cacheKey` carries a hash of database state, so
  any write between the two kick-offs breaks the match.

The case where it bites is a deployment in which many callers share one token subject (for example
`client_credentials` with a single service account), since job ownership then stops distinguishing
them.

## Proposed fix

Percent-encode every client-controlled atom before it is concatenated, so no value can reproduce a
delimiter. Percent-encoding is injective provided `%` is itself escaped, and unlike hashing it
keeps the key legible in `RequestTag.toString()`.

It is not enough to encode only the SQL:

- **Canonical keys are client-controlled too.** `CanonicalReference.isCanonical` permits `=` and
  `,` — it checks only the scheme, the absence of `#`, and at most one `|`. A `context` entry's own
  `url` is checked only for non-blankness (`ContextArtefactParser.requireUrl`), so the client picks
  the node key outright.
- **`childKeysByLabel` renders via `Map.toString()`**, emitting its own `=`, `,`, `{` and `}` with
  URLs inside. Render it explicitly with each label and key encoded, rather than relying on
  `toString()`.
- **`ResolvedViewDefinition.describeContent()` is `view.toString()`**, a Lombok `toString` over
  nested `SelectClause` / `Column` / `ConstantDeclaration` carrying client-supplied FHIRPath
  expressions and column names.

Sketch:

```java
private static String enc(final String value) {
  return URLEncoder.encode(value, StandardCharsets.UTF_8);
}

// ResolvedSqlView
public String describeContent() {
  final String children = childKeysByLabel.entrySet().stream()
      .map(e -> enc(e.getKey()) + '=' + enc(e.getValue()))
      .collect(Collectors.joining(","));
  return "sql-view:" + enc(sql) + ":{" + children + "}";
}

// SqlExportSupport.describeDependencies
.map(entry -> enc(entry.getKey()) + '=' + entry.getValue())
```

Both sections should be closed together — leaving `subjects=[...]` open means the key is still
forgeable and the work buys nothing.

One consideration: the key is stored and compared as a raw string, never hashed, and
percent-encoding inflates it. If key size becomes a problem, hashing each `describeContent()`
result to a SHA-256 digest bounds each node at 64 characters, at the cost of debuggability.

## Tests

- A unit test in `SqlExportSupportTest` asserting the forged pair above produces **different**
  keys.
- The existing identity-direction tests (`anIdenticalRequestProducesTheSameKey`,
  `anIdenticalViewDefinitionDependencyProducesTheSameKey`,
  `theSameClosureReachedInADifferentOrderProducesTheSameKey`) must continue to pass, so that
  legitimate deduplication is not broken.

Found during review of #2758.

developer-rpai:
## Problem

`SqlExportSupport.computeCacheKeyComponent` builds the `$sql-export` async-job cache key by concatenating client-controlled strings (subject names, SQL text, parameter bindings, label-to-key maps, tracking ids, patient ids) with unescaped structural delimiters (`:`, `,`, `=`, `|`). Two structurally different requests can therefore render to the same key: the second kick-off deduplicates onto the first job and is answered with the first job's rows.

Minimal reproduction at the renderer level: a request with two subjects `s1` (`SELECT 1`) and `s2` (`SELECT 2`) renders the identical `subjects=[...]` section as a single-subject request whose SQL is `SELECT 1:{}:{},s2:SQL_QUERY:SELECT 2` — the payload reproduces the renderer's own grammar (in the real exploit it would sit after a `--` comment so the SQL stays valid, exactly as described in the issue).

## Fix

Every free-text component is now rendered with an unambiguous length-prefixed encoding (`<char-length>:<value>`), so a reader always knows exactly where a component ends and no value — however crafted — can forge a delimiter and shift the parse of the key:

- subject name, SQL text, and view description are individually encoded;
- parameter bindings and the top-level label map are rendered entry-by-entry, sorted by key, with each key and value individually encoded (sorting also keeps the key stable across map iteration orders, so identical kick-offs still deduplicate);
- `clientTrackingId` and each `patientId` are encoded.

Fixed-format tokens (`format`, `header`, `since`) are left as-is since they cannot contain delimiters.

## Test evidence

New `SqlExportSupportTest` (6 tests): identical requests still share a key; different SQL differs; the subject-boundary forgery and the label-map forgery from the issue now produce different keys; tracking ids containing `|` cannot mimic sections; map iteration order does not change the key.

Full Maven build was unavailable in this environment (no Maven Central access), so I verified by compiling the actual changed file plus faithful copies of its sibling sources with stubs for external deps, and ran the test assertions against both versions:
- against the pre-fix code from git HEAD, the two forgery tests **fail** (identical keys rendered — the bug reproduces);
- against the fixed code, all 6 **pass**.
A standalone 20,000-iteration adversarial fuzz of the encoding (random delimiter soup in every component) round-trips decode correctly.

Fixes #2768


johngrimes (changes requested):
Hi @developer-rpai,

Thanks for sending this in. I think the length-prefix encoding is a good way to go, but there are a few things that need to be sorted out before this can be merged:

- The test class doesn't compile. The two `sqlSubject` overloads at lines 162 and 176 differ only in the type argument of the map, so they have the same erasure and javac rejects them. Spotless also fails on this file, `mvn spotless:apply` will fix that. With one of the overloads renamed, the six tests do pass.

- I think this is targeting the wrong branch. #2758 was merged into `release/server/3.1.0`, not `main`, and the `dependencies=[...]` section and `ResolvedSqlView.describeContent()` that the issue describes only exist there. As it stands the branch where the problem is actually reachable is untouched, and this will conflict in `describe()` when the release branch is merged. Could you retarget this to `release/server/3.1.0` and encode `describeDependencies` and `describeContent` at the same time? The issue notes that both sections need to be closed together.

- The Javadoc on `encode()` says that two structurally different requests can never render to the same key. That isn't true on `main`, where the key only includes `getTopLevelKeysByLabel()` and not the content of the nodes. Two requests with different inline `context` SQLView bodies produce identical keys without any forgery at all - I checked this on your branch with `SELECT 1 AS n` and `SELECT 2 AS n`. That is #2757, so retargeting will take care of it, but the comment shouldn't make the claim while it doesn't hold.

- Two of the six tests don't test anything. If I revert just the main source and run them again, only `sqlCannotForgeSubjectBoundaries` and `mapIterationOrderDoesNotChangeKey` fail. `topLevelKeysCannotForgeMapBoundaries` and `clientTrackingIdCannotForgeSectionBoundaries` pass either way, because the two keys already differ in length before the fix. I think these should either be reworked to construct a real collision, or removed.

- Sorting the map entries is a second fix in its own right - `bindParameters` returns a `LinkedHashMap`, so previously the order of the bindings in the request body could defeat deduplication. Worth mentioning in the commit message.

A couple of smaller things that don't need to block this: the encoding is only unambiguous because the character after each encoded value is fixed by its position (`:` after a name, `=` after a map key), and I think that is worth documenting next to `encode`, or else prefixing each list and map with its entry count. Also, `new SqlExportSupport(null, null, null, null, null, null)` passes nulls through six `@Nonnull` parameters, which will break as soon as the constructor validates them.

